PR #150, opened 2020-04-21
[Test Categories] Add Resilience Tests category

petorre:
Proposing separate Resilience Tests category to measure how much can containerized workloads run in shared cloud environment or still require their slice of carrier grade platform.


taylor:
@petorre we think we would like to have resilience added to a list of capabilities/properties which could be applied to multiple categories.  We support the concept of tagging a test and adding capabilities/properties such as resilience would be a good tag.

Issue #183, opened 2020-04-29
[CNFs] Create CoreDNS CNF example in example-cnfs folder, add README, add cnf-conformance.yml, add public helm chart support


lixuna:
[Acceptance Criteria]https://github.com/cncf/cnf-conformance/issues/183#issuecomment-625003752)

### [CNFs] Update CoreDNS example CNF to move to example-cnfs folder, add README and cnf-conformance.yml

Documentation tasks:
- [x]  create folder for CoreDNS in example-cnfs folder
- [x]  create [README for CoreDNS CNF], based on [Envoy example](https://github.com/cncf/cnf-conformance/blob/master/example-cnfs/envoy/README.md) (example: https://github.com/cncf/cnf-conformance/blob/master/sample-cnfs/sample-coredns-cnf/README.md)

Implementation tasks: TBD
- [x] create [cnf-conformance.yml], based on [Envoy example](https://github.com/cncf/cnf-conformance/blob/master/example-cnfs/envoy/cnf-conformance.yml) 
- [x] add support for public helm chart re: #169 
- [ ] Tag 1 or more people to peer review

CNF Testing task:
- [ ]  Follow the new example CNF's README.md to test the CNF
- [ ]  Add the Final Score in a comment in this ticket
- [ ]  Add the entire test result output in a comment in this ticket

**QA tasks**

Dev Review: 
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Needs Peer Review` column  
   - [ ] create Pull Request and follow check list
   - [ ] Assign 1 or more people for peer review
- [ ] if no, document what additional tasks will be needed

Peer review:
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Reviewer Approved` column
   - [ ] Approve pull request
- [ ] if no, 
   - [ ] document what did not go as expected, including error messages and screenshots (if possible)
   - [ ] Add comment to pull request
   - [ ] request changes to pull request

taylor:
Acceptance criteria: steps to test during development and peer review
---

- [ ]  I can follow the set up instructions as a `CNF Developer` https://github.com/cncf/cnf-conformance/blob/master/INSTALL.md#cnf-developer-install-and-usage-guide
- [ ]  I should see a CoreDNS CNF example folder
    - [ ]   Ie. in an example-cnfs folder with each example CNF having it’s own sub-folder with a config
[...]

PR #185, opened 2020-04-29
Adds CoreDNS as an example CNF

taylor:
# CoreDNS as example CNF

## Description:

Adds CoreDNS as an example CNF with working cnf-conformance.yml and README on usage and description.

## Issues:
- #183


lixuna:
Creating a new issue for "Need a yml validator that fails these config files. it willl save us from heartache that we got with CNCF.ci yml. The docker_deploy attribute is missing above."

PR #281, opened 2020-06-23
Logging Refactor

williscool:
# CNF Conformance PR Template

## Description
(name of issue/change)

## Issues:
Refs: #280 

## How has this been tested:
 - [ ] Covered by existing integration testing
 - [ ] Added integration testing to cover
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] Shared Packet K8s cluster
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] No updates required.

**Code Review**
- [ ] Does the test handle fatal exceptions, ie. rescue block

**Issue**
- [ ] Tasks in issue are checked off

Issue #280, opened 2020-06-22
[Refactor] Improve output by refactoring all console messaging to use standard logging


taylor:
[Acceptance Criteria](https://github.com/cncf/cnf-conformance/issues/280#issuecomment-647630612) 

# [Refactor] Improve output by refactoring all console messaging to use standard logging

**Short description:**
- refactor all output messages to use standard logging throughout code

Currently there is some messaging that is sent to the console (standard out) bypassing the logging functionality.  This means it is more difficult to show what is on the screen and it can become noise.  It also limits the ability to send the output to different places (eg. a log file or service)  

## TODO

**Tasks:**

- [x] Update all code providing developer log messages to use built-in logging functionality.
- [x] Update all code providing verbose end user messages to use custom verbose function and use verbose flagging to enable/disable
- [x] Choose appropriate logging and verbose level for "converted" code.
- [x] Set the default logging level for end-users (eg. info, fatal, debug) to be used by the binary
   - Does not create config.yml
   - No config.yml by default
   - What is the log level?  Is it set at compile or runtime?
- [x] Set the default logging level for end-users (eg. info, fatal, debug) building from source
   - Currently log level is set in [config.yml](https://github.com/cncf/cnf-conformance/blob/224c9e2f70ff3f46acec030a3c84cacb14c8d700/config.yml) which is checked into the repository
- [x] Update docs for contributing developers to the use of logging functions
- [x] Update docs for contributing developers to the use of verbose functions
- [x] Document how end users running tests can use verbose flag and what it does
- [x] Document how end users running tests can use set logging levels, what the defaults are and what they will show
- [x] DOCUMENT the precedence of setting logging level
   1. Cli flag is highest precedence
   2. Environment var is next level of precedence
   3. Config file is last level of precedence 
- [x] Spec tests covering setting different logging levels and overrides
- [x] Spec tests covering verbosity usage

**QA tasks**

Dev Review: 
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Needs Peer Review` column  
   - [ ] create Pull Request and follow check list
   - [ ] Assign 1 or more people for peer review
- [ ] if no, document what additional tasks will be needed

Peer review:
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Reviewer Approved` column
   - [ ] Approve pull request
- [ ] if no, 
[...]

PR #291, opened 2020-06-25
logging refactor part 2

williscool:
# logging refactor part 2  for #280 and #281 

## Description

refactoring logging part 2

always have message with them so puts are in function

## Issues:
Refs: #280 

also #281 

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/spec/cnf_conformance_spec.cr
+++ b/spec/cnf_conformance_spec.cr
@@ -4,8 +4,8 @@ require "colorize"
 
 describe CnfConformance do
   before_all do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
     `./cnf-conformance samples_cleanup`
     $?.success?.should be_true
     `./cnf-conformance setup`
@@ -15,20 +15,20 @@ describe CnfConformance do
   end
 
   after_all do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
     `./cnf-conformance samples_cleanup`
     $?.success?.should be_true
   end
 
   it "'all' should run the whole test suite", tags: "happy-path" do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
     # Test the binary
     # Build should already be present
     # build_s = `crystal build src/cnf-conformance.cr`
     # $?.success?.should be_true
-    # puts build_s 
+    # LOGGING.debug build_s 
     `./cnf-conformance samples_cleanup`
     response_s = `./cnf-conformance all cnf-config=./sample-cnfs/sample-coredns-cnf/cnf-conformance.yml`
     puts response_s
@@ -44,8 +44,8 @@ describe CnfConformance do
   end
 
   it "'scalability' should run all of the scalability tests", tags: "happy-path"  do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
     `./cnf-conformance samples_cleanup`
     response_s = `./cnf-conformance setup`
     puts response_s
--- a/spec/cnf_conformance_spec.cr
+++ b/spec/cnf_conformance_spec.cr
@@ -31,7 +31,7 @@ describe CnfConformance do
     # LOGGING.debug build_s 
     `./cnf-conformance samples_cleanup`
     response_s = `./cnf-conformance all cnf-config=./sample-cnfs/sample-coredns-cnf/cnf-conformance.yml`
-    puts response_s
+    LOGGING.info response_s
     $?.success?.should be_true
     (/PASSED: Helm readiness probe found/ =~ response_s).should_not be_nil
     (/PASSED: Helm liveness probe/ =~ response_s).should_not be_nil
@@ -48,11 +48,11 @@ describe CnfConformance do
     # LOGGING.debug `echo $KUBECONFIG`
     `./cnf-conformance samples_cleanup`
     response_s = `./cnf-conformance setup`
-    puts response_s
+    LOGGING.info response_s
     `./cnf-conformance sample_coredns_with_wait_setup`
     $?.success?.should be_true
     response_s = `./cnf-conformance scalability`
-    puts response_s
+    LOGGING.info response_s
     $?.success?.should be_true
     (/PASSED: Replicas increased to 3/ =~ response_s).should_not be_nil
     (/PASSED: Replicas decreased to 1/ =~ response_s).should_not be_nil
--- a/spec/configuration_lifecycle_spec.cr
+++ b/spec/configuration_lifecycle_spec.cr
@@ -3,8 +3,8 @@ require "colorize"
 
 describe CnfConformance do
   before_all do
-    # LOGGING.debug `pwd` 
-    # LOGGING.debug `echo $KUBECONFIG`
+    LOGGING.debug `pwd` 
+    LOGGING.debug `echo $KUBECONFIG`
 
     `./cnf-conformance samples_cleanup`
     $?.success?.should be_true
--- a/spec/configuration_lifecycle_spec.cr
+++ b/spec/configuration_lifecycle_spec.cr
@@ -3,8 +3,8 @@ require "colorize"
 
 describe CnfConformance do
   before_all do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
 
     `./cnf-conformance samples_cleanup`
     $?.success?.should be_true
--- a/spec/configuration_lifecycle_spec.cr
+++ b/spec/configuration_lifecycle_spec.cr
@@ -18,7 +18,7 @@ describe CnfConformance do
       `./cnf-conformance sample_coredns_source_setup verbose`
       $?.success?.should be_true
       response_s = `./cnf-conformance ip_addresses verbose`
-      puts response_s
+      LOGGING.info response_s
       $?.success?.should be_true
       (/FAILURE: IP addresses found/ =~ response_s).should_not be_nil
     ensure
@@ -30,7 +30,7 @@ describe CnfConformance do
       `./cnf-conformance sample_coredns`
       $?.success?.should be_true
       response_s = `./cnf-conformance liveness verbose`
-      puts response_s
+      LOGGING.info response_s
       $?.success?.should be_true
       (/PASSED: Helm liveness probe/ =~ response_s).should_not be_nil
     ensure
@@ -42,7 +42,7 @@ describe CnfConformance do
       `./cnf-conformance sample_coredns_bad_liveness`
       $?.success?.should be_true
       response_s = `./cnf-conformance liveness verbose`
-      puts response_s
+      LOGGING.info response_s
       $?.success?.should be_true
       (/FAILURE: No livenessProbe found/ =~ response_s).should_not be_nil
     ensure
@@ -54,7 +54,7 @@ describe CnfConformance do
       `./cnf-conformance sample_coredns`
       $?.success?.should be_true
       response_s = `./cnf-conformance readiness verbose`
-      puts response_s
+      LOGGING.info response_s
       $?.success?.should be_true
       (/PASSED: Helm readiness probe/ =~ response_s).should_not be_nil
     ensure
@@ -66,7 +66,7 @@ describe CnfConformance do
       `./cnf-conformance sample_coredns_bad_liveness`
       $?.success?.should be_true
       response_s = `./cnf-conformance readiness verbose`
-      puts response_s
+      LOGGING.info response_s
       $?.success?.should be_true
       (/FAILURE: No readinessProbe found/ =~ response_s).should_not be_nil
     ensure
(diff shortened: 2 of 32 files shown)


williscool:
at this point the puts that remain im pretty sure we want always go to command line and not (only? should we do both?) in `LOGGING` 

these cases include

- test score outputs
- fatal error messages
- task sucess / fail messages

spec/cnf_conformance_spec.cr
```diff
@@ -15,23 +15,23 @@ describe CnfConformance do
...
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
     # Test the binary
     # Build should already be present
     # build_s = `crystal build src/cnf-conformance.cr`
     # $?.success?.should be_true
-    # puts build_s 
+    # LOGGING.debug build_s
```
taylor, line 31: @williscool Now that these are LOGGING.debug level do you want to uncomment them?

spec/configuration_lifecycle_spec.cr
```diff
@@ -3,8 +3,8 @@ require "colorize"
 
 describe CnfConformance do
   before_all do
-    # puts `pwd` 
-    # puts `echo $KUBECONFIG`
+    # LOGGING.debug `pwd` 
+    # LOGGING.debug `echo $KUBECONFIG`
```
taylor, line 7: Uncomment?

spec/utils/sample_utils_spec.cr
```diff
@@ -146,12 +146,12 @@ it "'validate_cnf_conformance_yml' (function) should warn, but be valid when a c
...
 
   yml = parsed_config_file(ensure_cnf_conformance_yml_path(args.named["cnf-config"].as(String)))
-  puts yml.inspect
+  LOGGING.info yml.inspect
   ("#{yml.get("release_name").as_s?}").should eq("coredns")
 
   status, warning_output = validate_cnf_conformance_yml(yml)
 
-  puts "WARNING: #{warning_output}"
+  LOGGING.info "WARNING: #{warning_output}"
```
taylor, line 154: How about LOGGING.warning?

spec/utils/utils_spec.cr
```diff
@@ -209,13 +209,13 @@ describe "Utils" do
...
       privileged_list = privileged_response.to_s.split(" ").uniq
-      puts "privileged_list #{privileged_list}"
+      LOGGING.info "privileged_list #{privileged_list}"
       if privileged_list.select {|x| x == helm_chart_container_name}.size > 0
         resp = "✖️  FAILURE: Found privileged containers: #{privileged_list.inspect}".colorize(:red)
       else
         resp = "✔️  PASSED: No privileged containers".colorize(:green)
       end
-      puts resp
+      LOGGING.info resp
```
taylor, line 218: This looks like the end-user output.  Should this use the new command for sending the output for PASSED/FAILED to the user?

spec/utils/utils_spec.cr
```diff
@@ -209,13 +209,13 @@ describe "Utils" do
...
       resp
     end
     (task_response).should eq(["✔️  PASSED: No privileged containers".colorize(:green), "✔️  PASSED: No privileged containers".colorize(:green)])
```
taylor, line 221: What happens to the spec when this is switch to LOGGING.info?  Do we need to ensure LOGGING level is set to something specific for specs to pass?

Maybe we should not use LOGGING in the specs?

spec/utils/utils_spec.cr
```diff
@@ -235,13 +235,13 @@ describe "Utils" do
...
       if privileged_list.select {|x| x == helm_chart_container_name}.size > 0
         resp = "✖️  FAILURE: Found privileged containers: #{privileged_list.inspect}".colorize(:red)
       else
         resp = "✔️  PASSED: No privileged containers".colorize(:green)
       end
-      puts resp
+      LOGGING.info resp
       resp
     end
     (task_response).should eq("✔️  PASSED: No privileged containers".colorize(:green))
```
taylor, line 247: Maybe we should not use LOGGING in the specs?  See previous comments

src/cnf-conformance.cr
```diff
@@ -8,19 +8,19 @@ task "all", ["all_prereqs", "configuration_file_setup", "compatibility","statele
...
-    puts "Conformance Suite failed!".colorize(:red)
-    puts "Failed required tasks: #{failed_required_tasks.inspect}".colorize(:red)
+    stdout_failure "Conformance Suite failed!"
+    stdout_failure "Failed required tasks: #{failed_required_tasks.inspect}"
   end
 
   # new_results = create_final_results_yml_name
   # results = `mv #{LOGFILE} #{new_results}`
-  puts "Results have been saved to #{Results.file}"
+  stdout_info "Results have been saved to #{Results.file}"
```
taylor, line 23: Why is it called stdout_*?  Could this be a more general human action vs C-based "word"

spec/utils/utils_spec.cr
taylor: Or specs should use the new command (eg. stdout_success/stdout_failure)

src/tasks/configuration_lifecycle.cr
```diff
@@ -138,8 +133,9 @@ task "rolling_update" do |_, args|
     end
     
     unless version_tag
-      upsert_failed_task("rolling_update")
-      raise "✖️  FAILURE: please specify a version of the CNF's release's image with the option version_tag or with cnf_conformance_yml option 'rolling_update_tag'"
+      fail_msg = "✖️  FAILURE: please specify a version of the CNF's release's image with the option version_tag or with cnf_conformance_yml option 'rolling_update_tag'"
+      upsert_failed_task("rolling_update", fail_msg)
+      raise fail_msg
```
taylor, line 138: Is this a hard failure causing the test to exit?  It's pretty confusing to have stdout_failure used for both PASSED/FAILED and hard failures, then have a completely different way of having failure messages.

`stdout_failure "Error: You must supply either cnf-config or cnf-path".colorize(:red)`

src/tasks/installability.cr
```diff
@@ -7,12 +7,7 @@ require "./utils/utils.cr"
 
 desc "The CNF conformance suite checks to see if CNFs support horizontal scaling (across multiple machines) and vertical scaling (between sizes of machines) by using the native K8s kubectl"
 task "installability", ["install_script_helm", "helm_chart_valid", "helm_chart_published", "helm_deploy"] do |_, args|
-  total = total_points("installability")
-  if total > 0
-    puts "Installability final score: #{total} of #{total_max_points("installability")}".colorize(:green)
-  else
-    puts "Installability final score: #{total} of #{total_max_points("installability")}".colorize(:red)
-  end
+  stdout_score("installability")
```
taylor, line 10: Suggest not using "stdout" in the name

src/tasks/installability.cr
```diff
@@ -122,19 +112,15 @@ task "helm_chart_published", ["helm_local_install"] do |_, args|
...
         end
       else
-        upsert_failed_task("helm_chart_published")
-        puts "✖️  FAILURE: Published Helm Chart Not Found".colorize(:red)
+        upsert_failed_task("helm_chart_published", "✖️  FAILURE: Published Helm Chart Not Found")
       end
     else
-      upsert_failed_task("helm_chart_published")
-      puts "✖️  FAILURE: Published Helm Chart Not Found".colorize(:red)
+      upsert_failed_task("helm_chart_published", "✖️  FAILURE: Published Helm Chart Not Found")
```
taylor, line 123: A second way of showing PASSED/FAILED

src/tasks/utils/system_information/helm.cr
```diff
@@ -6,61 +6,61 @@ def helm_installation(verbose=false)
...
-    puts "Helm not found".colorize(:red)
-    puts %Q(
+    stdout_failure "Helm not found"
+    stdout_failure %Q(
     Installation instructions for Helm can be found here: https://helm.sh/docs/intro/install
 
     To install helm on Linux use:
     sudo snap install helm --classic
-    ).colorize(:red)
+    )
```
taylor, line 40: Nice that quoted messages work 👍

src/tasks/utils/utils.cr
```diff
@@ -459,5 +452,30 @@ def results_by_tag(tag)
...
+def stdout_score(test_name)
+  total = total_points(test_name)
+  pretty_test_name = test_name.split("_").map(&.capitalize).join(" ")
+  test_log_msg = "#{pretty_test_name} final score: #{total} of #{total_max_points(test_name)}"
+
+  if total > 0
+    stdout_success test_log_msg
+  else
+    stdout_failure test_log_msg
+  end
```
taylor, line 480: If possible have this work/merge with the upsert_PASSED/FAILED_task code.  eg.

`upsert_passed_task("privileged", "✔️  PASSED: No privileged containers")`

taylor:
> do

taylor:
> at this point the puts that remain im pretty sure we want always go to command line and not (only? should we do both?) in `LOGGING`
> 
> these cases include
> 
> * test score outputs
> * fatal error messages
> * task sucess / fail messages

Yes.  Some type of wrapper for those messaging (what you have done or the upsert_passed_task style) to differentiate from LOGGIN sounds good.  We may *also* want to call the LOGGING.

spec/cnf_conformance_spec.cr
williscool: @taylor that is the question my dear watson lol.

I thought about this myself but wanted other inputs.

williscool: I don't think all of them need to be. I think people can drive by do that over time where needed as needed.

for instance this commented code could probably be deleted all together because its doing a crystal build which is already done by the spec helper now

spec/configuration_lifecycle_spec.cr
williscool: why not

spec/utils/sample_utils_spec.cr
williscool: why not

spec/utils/utils_spec.cr
williscool: so I think the answer to this is yes and no.

No because this is a test so the output going to stdout here isn't what we want (becasue the assertion is on the value returned from the `task_response` function

.... 

But yes because what that function should do is be inside of an actual task we would want stdout for... and then invoked inside of this test.

so I would leave this for now

src/tasks/installability.cr
williscool: I used that name to be explicit about the fact that this is output we always want to go to stdout regarless of whether or not its in LOGGING

any other suggestions?

spec/utils/utils_spec.cr
williscool: so I think the answer to this is yes and no.

No because this is a test so the output going to stdout here isn't what we want (becasue the assertion is on the value returned from the task_response function

....

But yes because what that function should do is be inside of an actual task we would want stdout for... and then invoked inside of this test.

so I would leave this for now

spec/utils/utils_spec.cr
williscool: I think we definitely want LOGGING almost always in specs.

This will help cut down on spurious output and make logs easier to debug with since you'll be able to set log level to what you want and ignore lots of stuff you don't care about

src/tasks/configuration_lifecycle.cr
williscool: I'm confused by what you mean by " It's pretty confusing to have stdout_failure used for both PASSED/FAILED" there shouldn't be any places where `stdout_failure` is used and the test succedes or vice versus

src/tasks/utils/utils.cr
williscool: dont really see a why to merge these the code that calculates the scores is pretty seperate

williscool:
I think I understand your confusion taylor

`stdout_*` are **purely** presentation functions intended to be exactly like logging but always go to stdout. 

They are designed to centralize the presentation of all output we always want on the command line and and explicitly NEVER disabled or disableable. Simiarly to the way that `LOGGING` is designed to be able to be turned off with a log level.


the `upsert_*_task` functions on the other hand have a side effect of updating the points for a task. they use the presentational `stdout_` functions interally to output the success or failure msgs to the console

spec/utils/utils_spec.cr
williscool: Also to answer you other question about "Do we need to ensure LOGGING level is set to something specific for specs to pass?"

The answer is no. all of the places in tests currently where logging is used it is to show command line output of a task invoked with backticks and can be safely disabled

Would defintely be an anti pattern to need a specific log level for tests to pass.

src/tasks/configuration_lifecycle.cr
taylor: I see the CNF test FAILED message and critical failure for an internal test for cnf-config option missing using stdout_failure.  I also see places were a CNF test FAILED uses upsert_failed_task instead of stdout_failed.

src/tasks/installability.cr
taylor: Hmmm I would hide the standard out in the main function used.  What we care about is communicating to the user running the test if the tests PASSED/FAILED and what the scores are.  Maybe that is via stdout on CLI.  Maybe it is via SMS response ;)  

LOGGING tells me it's about saving the output somewhere.  These messages are for the interactive end user or maybe test results. 

Not sure about naming... 
- test_passed()
- test_failed()
- test_scores()

williscool:
merging now. will address comments in new pr per msg w @taylor

PR #362, opened 2020-08-13
make wait_for_install wait on all replicas

clementnuss:
## Description

Without this fix, wait_for_install waits for at least one replica to be ready. However for a deployment of a CNF with multiple replicas, we need to wait for all replicas to be in the ready state before declaring it 'installed'.
At least this is what is assumed by the `chaos_container_kill` test, which uses the `wait_for_install` method to check that after a pod is killed, a new one is spawned to match the number of desired replicas.
In the previous version of the function, when there are more than 1 replica (e.g. with the `envoy` example), the test does not wait and produces a wrong result.

## How has this been tested:
 - [ ] Test environment
    * [x] Bare-metal K8s cluster

## Types of changes:

**Code Review**
- [ ] ⚠ Is it ok to make `wait_for_install` wait for all replicas to be ready ? what side effects could this have ?

### DIFF
--- a/src/tasks/utils/sample_utils.cr
+++ b/src/tasks/utils/sample_utils.cr
@@ -48,9 +48,10 @@ end
 def wait_for_install(deployment_name, wait_count=180, namespace="default")
   second_count = 0
   all_deployments = `kubectl get deployments --namespace=#{namespace}`
+  desired_replicas = `kubectl get deployments --namespace=#{namespace} #{deployment_name} -o=jsonpath='{.status.replicas}'`
   current_replicas = `kubectl get deployments --namespace=#{namespace} #{deployment_name} -o=jsonpath='{.status.readyReplicas}'`
   LOGGING.info(all_deployments)
-  until (current_replicas.empty? != true && current_replicas.to_i > 0) || second_count > wait_count.to_i
+  until (current_replicas.empty? != true && current_replicas.to_i == desired_replicas.to_i) || second_count > wait_count.to_i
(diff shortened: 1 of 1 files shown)


clementnuss:
Thank you @williscool for the `ReleaseManager` changes, I can now run the specs without errors 👍

williscool:
👍 @clementnuss you should be able to rebase and push and pass now then we can merge

clementnuss:
Just tried it, however I get the following result:
```
Finished in 17:11 minutes
117 examples, 0 failures, 0 errors, 0 pending
Error running at_exit handler: Missing ENV key: "GITHUB_USER"
```

Any idea why ?
Also, I still have some of the log deprecation warnings, I was hoping your latest PR would have solved these, it is still in progress ? (Good luck with that, refactoring a logging solution is time-consuming 😅😬

williscool:
hmm .. it did solve that issue for my other PRs last week..

Will dig into why it's not working on yours here

clementnuss:
Maybe a precision, the message I put in my last comment is the output of the Travis CI for my latest push (with your changes), when I run the code locally all tests do pass correctly.

denverwilliams:
@clementnuss @williscool I went ahead and merged because it's passing all the spec tests now. Travis was exiting because it was still trying to publish a release for the PR, which it needs the github keys to do. 

There is a PR [here](https://github.com/cncf/cnf-conformance/pull/364) which will disable automatic publishing for external PRs, so once it's merged you shouldn't run into this again.

Issue #422, opened 2020-10-01
[BUG] Runner needs to clean up jobs, even when the build fails


wavell:
**Describe the bug**

Packet runner needs to have its kind clusters clean up **even when the build fails**
Here is the error that it gives after too many kind clusters appear:
```
Error from server (NotFound): error when deleting "/cnf-conformance/spec/fixtures/chaos_network_loss.yml": networkchaos.pingcap.com "network-loss-test" not found
Error from server (NotFound): error when deleting "/cnf-conformance/spec/fixtures/chaos_cpu_hog.yml": stresschaos.pingcap.com "cpu-hog-test" not found
The job exceeded the maximum time limit for jobs, and has been terminated.
```

**To Reproduce**
1. Build on travis where the builds fail repeatedly

**Expected behavior**
No build errors

williscool:
LGTM 

failure_func cleans jobs after a failed job

https://travis-ci.org/github/cncf/cnf-conformance/builds/733198313

PR #423, opened 2020-10-01
Fix bug where cleanup doesn't run when the Travis job fails. #422

denverwilliams:
## Description
Fix bug where cleanup doesn't run when the Travis job fails.

## Issues:
Refs: cncf/cnf-conformance/issues/422

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ jobs:
           if [ -z "${SSH_KEY+x}" ]; then
             echo "SSH Key not found, using Travis Runner"
           # Failure Function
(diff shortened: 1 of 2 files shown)


taylor (approved):
@denverwilliams  Suggestion: use bash trap function to call the failure_func on all exits.  

Example

`trap failure_func EXIT`

You can trap for other signals to if it's possible they would happen and should be caught 

After trap is used you can do exit 1 after the || 

Also I recommend moving the failure func higher up instead of in the else block so it's global and always tries a cleanup.

PR #452, opened 2020-10-30
Update README.md

xmulligan:
Updates to introducing the CNF Conformance program


taylor (commented):
Maybe the interoperability benefit could be rewritten and possibly moved so it's more explicitly communicating it as a end benefit of CNFs which are following cloud native methodologies.

taylor (approved):
lgtm

Issue #269, opened 2020-06-11
Statelessness test: Does the CNF use a non-cloud native data store: local volumes on the node?


lixuna:
[Acceptance Criteria](https://github.com/cncf/cnf-conformance/issues/269#issuecomment-645545163)
---

### Statelessness test: Does the CNF use a non-cloud native data store: local volumes on the node?

**Short Description:** 
- Check for non-cloud native data stores
- Does the CNF use local block volumes on the node?

Reference:
- https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga
- https://kubernetes.io/docs/concepts/storage/volumes/#local

Raw block volumes to be explored in another test:
- https://kubernetes.io/docs/concepts/storage/volumes/#csi-raw-block-volume-support
- https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support

**Test Category:** 
- Statelessness (eg. Cloud native State) 

**Type of test (static or runtime)**
- Runtime

**Scoring:** 
- tbd

**Are there any pre-requisites?**
- tbd

**Command name:**
- tbd

---

### Implementation Tasks: TBD

**Environment set up tasks:**
- [ ]  if needed

**Upstream tool set up tasks: (conformance suite + upstream tools)**
- [ ]  if needed

**Code implementation tasks:**
- [ ]  update pre-req checker - N/A

**Documentation tasks:**
- [ ] Update [Test Categories md](https://github.com/cncf/cnf-conformance/blob/master/TEST-CATEGORIES.md) if needed 
- [ ] Update [USAGE md](https://github.com/cncf/cnf-conformance/blob/master/USAGE.md) if needed 
- [ ] Update [installation instructions](https://github.com/cncf/cnf-conformance#installation)  if needed 

### QA tasks

Dev Review: 
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Needs Peer Review` column  
   - [ ] create Pull Request and follow check list
   - [ ] Assign 1 or more people for peer review
- [ ] if no, document what additional tasks will be needed

Peer review:
- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes, 
   - [ ] move to `Reviewer Approved` column
   - [ ] Approve pull request
- [ ] if no, 
   - [ ] document what did not go as expected, including error messages and screenshots (if possible)
   - [ ] Add comment to pull request
[...]

PR #468, opened 2020-11-11
Statelessness test: Does the CNF use a non-cloud native data store: local volumes on the node? #269

wavell:
# Statelessness test: Does the CNF use a non-cloud native data store: local volumes on the node? #269

## Description
Statelessness test: Does the CNF use a non-cloud native data store: local volumes on the node? #269

## Issues:
Refs: #269

## How has this been tested:
 - [x] Added integration testing to cover
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] I have updated the documentation accordingly.
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
+++ b/sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
@@ -73,17 +73,13 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         volumeMounts:
-        - name: state-volume
+        - name: task-pv-storage
           mountPath: /home
         resources:
 {{ toYaml .Values.resources | indent 10 }}
         ports:
 {{ include "coredns.containerPorts" . | indent 8 }}
       volumes:
-        - name: state-volume
-          hostPath:
-            path: /tmp/{{ randAlphaNum 10 }}
-            type: DirectoryOrCreate	
         - name: task-pv-storage
           persistentVolumeClaim:
             claimName: foo-pvc 
--- a/sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
+++ b/sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
@@ -1,0 +1,89 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "coredns.fullname" . }}
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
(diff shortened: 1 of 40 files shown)


sample-cnfs/sample-local-storage/chart/templates/persistent-volume.yaml
```diff
@@ -0,0 +1,23 @@
...
+  local:
+    path: /var/tmp
+  nodeAffinity:
+    required:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: kubernetes.io/hostname
+          operator: In
+          values:
+          - kind-control-plane
```
taylor, line 22: @denverwilliams @wavell FYI, this test will not work on a non-kind cluster (unless someone happens to name one of their nodes "kind-control-plane").

We probably need to look at another affinity item to use or add support for dynamically setting the node name.

sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+          mountPath: /home
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        ports:
+{{ include "coredns.containerPorts" . | indent 8 }}
+      volumes:
+        - name: state-volume
+          hostPath:
+            path: /tmp/{{ randAlphaNum 10 }}
+            type: DirectoryOrCreate
```
taylor, line 86: @wavell @denverwilliams This does not mean that are actively using a local volume to retrieve data that was written there in the past.  They could still be using a local volume for temporary storage.

taylor (approved):
left some comments.  looks like a good first version.

my main suggestion is about using a known path every time and checking for existing data as the test vs random.  But if it's past in as a template that could be random or the same every time. Not sure how to check.

Is it okay to use a local volume if it's only for ephemeral data?

agentpoyo (approved):
LGTM. :+1: 

As @taylor  pointed out, my only concern would be that it's only going to work on kind cluster.

sample-cnfs/sample-local-storage/chart/templates/persistent-volume.yaml
wavell: Good catch, we addressed this today

sample-cnfs/sample-local-storage/chart/templates/deployment.yaml
wavell: We aren't looking in the deployment yaml but in the persistent volume claim yaml.  We are looking for the existence of the local: and path: keys

denverwilliams: @taylor This hostPath code has nothing to do with the new test, it's a remnant from the old hostpath sample-cnf we used as a base. It's been removed it from the helm chart now.

PR #556, opened 2021-01-07
Link to releases/latest for binary download

taylor:
Current link goes to /releases which means the user will have to wade through all of the dev snapshots.  This update links to https://github.com/cncf/cnf-conformance/releases/latest for the latest stable release


agentpoyo (approved):
LGTM

PR #796, opened 2021-05-31
doc updates to reference mac os x support for source installs #647

agentpoyo:
## Description
MacOS source install verification and proof of concept. 

## Issues:
Refs: #647 

## How has this been tested:
 - [ ] Covered by existing integration testing
 - [ ] Added integration testing to cover
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] Shared Packet K8s cluster
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [x] Documentation update

## Checklist:
**Documentation**
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] No updates required.

**Code Review**
- [ ] Does the test handle fatal exceptions, ie. rescue block

**Issue**
- [ ] Tasks in issue are checked off

PR #802, opened 2021-06-03
move arch under platforms

taylor:
arch should be close to platform in the FAQ to make these related items easy to find.

ref #796


agentpoyo:
Updates were done in https://github.com/cncf/cnf-testsuite/pull/796

Issue #1239, opened 2022-02-28
[BUG] Handle helm "INSTALLATION FAILED" error


HashNuke:
## Describe the bug
When Helm install fails, Helm outputs an error message with a summary/reason for the failure.

Below are few examples of what the error lines might look like:
```
Error: INSTALLATION FAILED: failed to download "stable/prometheus-adapter"

Error: INSTALLATION FAILED: path "/github-runner-cnf-testsuite/cnf-testsuite/cnf-testsuite/cnfs/coredns/cnfs" not found

Error: INSTALLATION FAILED: parse error at (coredns/templates/deployment.yaml:16): function "sdfskfsdf" not defined
```

## Expected behaviour

Handle these errors in CNF Manager and other areas where `Helm.install` helper is being used.

agentpoyo:
###  Acceptance Criteria 
---
### Documentation is updated:
- [x] USAGE.md documentation updated
   - [x] how to run
   - [x] Description, remediation and details 
       - [x] what the best practice is
       - [x] why are we testing this
- [ ] CATEGORIES updated

###  Common steps
- [x]  Find or create a K8s cluster to use
   - Talk with team for access to an existing cluster
   - [Kind K8s clusters](https://github.com/cncf/cnf-testsuite/blob/main/KIND-INSTALL.md) will work as well
- [x]  Set the KUBECONFIG environment variable for accessing the target K8s cluster
- [ ]  As a developer, Pull down latest version of CNF conformance test code from {FEATURE} branch
     - [ ]  Check that {FEATURE} Branch passes all A/C 
     - [ ]  Check that {FEATURE} Branch passes tests in GitHub Actions
- [x]  As a peer reviewer, check that PR passes in GitHub Actions
     - [x] if yes, merge PR to master
     - [x]  if no, make a comment and request that developer make changes to pass tests 
- [x]   As a peer reviewer, Pull down latest version of CNF conformance test code from `main` branch
- [ ]  Run `shards install`

### Update Prerequisite checker
- [ ] NA

### Added to overall points:
- [ ]  I see the test in the embedded points.yml as valid test score

### When I run `cnf_setup` I should not see generic errors of helm installation failures:
- [ ] Copy `./sample-cnfs/sample_coredns/cnf-testsuite.yml' to your pwd and edit it to use a bogus helm chart.
- [ ] Run `./cnf-testsuite cnf_setup cnf-config=./cnf-testsuite.yml`
[...]

PR #1240, opened 2022-03-01
[1219, 1239] Moves third-party tools to cnf-testsuite namespace to avoid prevent test failures

### LINKED ISSUE #1219: [BUG] Third-party tools installed by testsuite are causing test failures
## Describe the bug

> This issue is an off-shoot from #1199. Some base work has already been done in #1199.

Third-party tools like falco, dockerd, etc installed by testsuite are causing test failures. Screenshots below show the test results for specific tests.

The tools that cause test failures are:
* falco
* dockerd
* chaos_operator
* local-path-provisioner
* Litmus Chaos
* pod-io-stress

### Test: `privilege_escalation`

image

### Test: `resource_policies`

image

### Test: `linux_hardening`

image

### Test: `ingress_egress_blocked`

image

### Test: `privileged_containers`

image

### Test: `network_policies`

image

### Test: `immutable_file_systems`

image

### Test: `hostpath_mounts`

image

## To reproduce

> To reproduce, run the workload tests twice.

The bug is not reproduce-able running individual tests. This is why our CI is less likely to catch these issues.

1. Run the workload tests using the below command to reproduce the same issues as the screenshots above.
2. The first run will not throw failures about dockerd and pod-stress-io. For this to show up as failures, run the workload tests for the second time.

```
crystal run src/cnf-testsuite.cr -- workload
```

## Expected behaviour

Tools installed by the testsuite should not cause test failures

## Proposed solution

* Third-party tools can be moved into a different namespace to avoid causing the issue described in this ticket.
* cluster-tools is already being moved to a `cnf-testsuite` namespace as a part of #1199.

HashNuke:
## Screenshots from workload tests to validate PR changes

> **IMPORTANT**: To validate the changes in this PR, please run the testsuite more than once.
>
> This is because the third-party tools interfering in tests happens more on the second run as per #1219. (Not uninstalling 3rd-party tools after using them is one of the reasons for #1219).

image

image

#### src/tasks/utils/kubescape.cr
* Define namespaces to skip in kubescape scan

#### src/tasks/dockerd_setup.cr
* Remove top-level dockerd helper functions. These have been moved to Dockerd module at `src/tasks/utils/dockerd.cr`
* Use new Dockerd helper functions to install and uninstall the tool in the cnf-testsuite namespace.

#### src/tasks/falco_setup.cr
* Install and uninstall Falco from cnf-testsuite namespace

#### src/tasks/fluentd_setup.cr
* Remove `FluentDManager` module and move all helpers to the existing FluentD module at `src/tasks/utils/fluentd.cr`
* Use those new helpers to install and uninstall.

#### .github/workflows/actions.yml
* Add insecure registry config to all kind cluster configs in GitHub Actions file

#### Set repository host in CNF to full cluster url in sample CNFs
The following files were updated:
* `sample-cnfs/sample_local_registry/chart/values.yaml`
* `sample-cnfs/sample_local_registry_org_image/chart/values.yaml`

#### spec/workload/installability_spec.cr
* After the tests use `cnf_cleanup` task instead of custom tasks to cleanup CNFs
* For the `'helm_chart_valid' should fail on a bad helm chart` test, after the test, run `cnf_cleanup` instead of `cnf_setup`

#### spec/workload/registry_spec.cr
* Run `setup` task in before_all to create `cnf-testsuite` namespace
* Use new `Dockerd` module helpers to work with `dockerd` pod and for install/uninstall of the resource.
* Use full cluster url for the private registry in the specs

#### spec/workload/security_spec.cr
* Since 1219 has been fixed, the test does not output a failure and the CNF is a good/passing CNF. The spec test has now been updated to reflect the positive scenario.
* Rename test
  * From this: `'resource_policies' should fail on a cnf that has containers with no resource limits defined`
  * To this: `'resource_policies' should pass on a cnf that has containers with resource limits defined`

#### src/tasks/utils/falco.cr
* Install Falco in the testsuite namespace

#### src/tasks/workload/compatibility.cr
* Increase timeout to 100s to avoid CI failures.

#### src/tasks/workload/microservice.cr
* Use new Dockerd module helpers
* Do not use `-i` flag for kubectl exec since we do not need interactive terminals. This removes the warnings in the CI output.

#### utils/kubectl_client/kubectl_client.cr
* Update `KubectlClient::Delete.command` to support label selectors
* Update `KubectlClient.logs` to support the options and namespace arguments.

#### src/tasks/workload/observability.cr
* Update call to `KubectlClient.logs` with named argument for `options`

#### src/tasks/workload/reliability.cr
* After Litmus test is complete, ensure to delete all the litmus resources using label selector.

#### src/tasks/workload/security.cr
* Look for Falco in the testsuite namespace

#### src/tasks/workload/security.cr
* `privileged_containers` does not have to uninstall & reinstall dockerd and cluster-tools anymore since they are in a new namespace.

#### tools/dockerd/manifest.yml
* Add new insecure registry to the dockerd manifest

#### utils/helm/helm.cr
`Helm.install` should raise new `Helm::InstallationFailed` when there is an installation error.

#### src/tasks/utils/cnf_manager.cr
* Rescue from the `Helm::InstallationFailed` error where required
* Define new `helm_error` var to record error state. This necessary to record the right `helm_used` value in the startup information configmap.

#### spec/platform/observability_spec.cr
* Updated spec to use `Helm.install`  and  `Helm.uninstall` helpers

#### spec/utils/release_manager_spec.cr
* Minor change when debugging release test. Use `eq` instead of `==`  followed by `be_truthy`. Helps with debugging output when value is not expected.

#### src/tasks/litmus_cleanup.cr
* Move `uninstall_litmus` task to `litmus_setup.cr` . Just consolidates both install and uninstall in one file. No other changes.
* Deletes `litmus_cleanup.cr` file

#### src/tasks/utils/opa.cr
* Do not have to convert `IO::Memory#to_s` because output is now a String (`KubectlClient.describe` has been updated)

#### src/tasks/utils/utils.cr
* `ShellCmd.run`  will only log stderr if there is an error. This avoids unnecessary stderr log lines in the CI output.

## Issues:
Refs: #1219, #1239

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


agentpoyo (approved):
LGTM

Issue #1431, opened 2022-05-04
[BUG] Update Usage Documentation for the cnf-testsuite.yml Generator


denverwilliams:
**Describe the bug**
Update usage documentation for the cnf-testsuite.yml generator & remove container_names for the generator as it's currently not needed in the cert.

**To Reproduce**

1. Go to [CNF_TESTSUITE_YML_USAGE.md](https://github.com/cncf/cnf-testsuite/blob/main/CNF_TESTSUITE_YML_USAGE.md)
2. Go to "[What is the cnf-testsuite.yml and why is it required?](https://github.com/cncf/cnf-testsuite/blob/main/CNF_TESTSUITE_YML_USAGE.md#what-is-the-cnf-testsuiteyml-and-why-is-it-required)"
4. Read information
5. See container name as part of the details needed.

**Expected behavior**
container names are not required so should not be mentioned

(optional: unnecessary for things like spelling errors and such)

Once this issue is address how will the fix be verified?

**Additional context**

--

taylor:
@denverwilliams documentation still mentions container names here https://github.com/cncf/cnf-testsuite/blob/main/CNF_TESTSUITE_YML_USAGE.md#what-is-the-cnf-testsuiteyml-and-why-is-it-required

taylor:
All good.  This can be closed.

From Denver:

> allowlist_helm_chart_container_names   is used for excluding privileged containers and entirely separate argument from container_names

PR #1455, opened 2022-05-10
Removed container required info from yml docs

taylor:
Minor update to yml docs for container names which is no longer required.  Simplified first sentences.

## Issues:
Refs: #1431


taylor:
keeping as is and closing as invalid

Issue #1684, opened 2022-10-12
[Workload] Check if the operator is installed with privileged rights.


wavell:
## Title: [Workload] Check if the operator is installed with privileged rights. 

**Is your workload test idea related to a problem? Please describe.**
- security

**Describe the solution you'd like**
- Check if the operator is installed with privileged rights.

**Test Category Name**
- Use existing Security category

**Type of test**
- runtime

---

### Documentation tasks:
- [ ] Update [installation instructions](https://github.com/cncf/cnf-testsuite/blob/main/install.md) if needed
- [ ] Update [Test Categories md](https://github.com/cncf/cnf-testsuite/blob/main/TEST-CATEGORIES.md) if needed
- [ ] Update [USAGE md](https://github.com/cncf/cnf-testsuite/blob/main/USAGE.md) if needed
  - [ ] How to run
  - [ ] Description and details
    - [ ] What the best practice is
    - [ ] Why are we testing this
  - [ ] Remediation steps if test does not pass

### QA tasks

Dev Review:

- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes,
  - [ ] move to `Needs Peer Review` column
  - [ ] create Pull Request and follow check list
  - [ ] Assign 1 or more people for peer review
- [ ] if no, document what additional tasks will be needed

Peer review:

- [ ] walk through A/C
- [ ] do you get the expected result?
- [ ] if yes,
  - [ ] move to `Reviewer Approved` column
  - [ ] Approve pull request
- [ ] if no,
  - [ ] document what did not go as expected, including error messages and screenshots (if possible)
  - [ ] Add comment to pull request
  - [ ] request changes to pull request

wavell:
points:
watson: 8
Denver: 8

lixuna:
@wavell @denverwilliams @taylor @HashNuke @agentpoyo can we close this issue and the related WIP PR? If not, what do you suggest as the next steps?

wavell:
It's a valid test.  Maybe someone else wants to take it over.

PR #1799, opened 2023-06-16
[WIP] Check if operator is installed with privileged rights

williscool:
## Description

title says it all

## Issues:
Refs:  https://github.com/cncf/cnf-testsuite/issues/1684

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] No updates required.

**Code Review**

**Issue**


martin-mat:
Closing this draft due to age, but explicitly *not* abandoning the feature — issue #1684 stays open and the test is still wanted. It also fills a real gap: today's security tests enumerate resources from the composite installed manifest, while an OLM operator's Deployment comes from the CSV at runtime, so operator pods are invisible to them.

The branch itself is no longer salvageable as code:
- It's ~650 commits behind; the task/result API (`task_runner` now yields a `result` with `passed/failed/na`), KubectlClient/Helm APIs, spec conventions (`ShellCmd.run_testsuite` + `verify_task_result`, mandatory tags), and config format (v2) have all changed since 2023.
- The setup pipeline was never finished — `generate_bundle`/`setup_catalog` are empty stubs and the fixture depends on a manually built `localhost:5001` catalog, so the failing-path spec could never run in CI.
- The core check compares a JSON boolean against the string `"true"`, so the test can never actually fail; it also digs the wrong CSV path (`install.deployments` vs `spec.install.spec.deployments`) and reports PASSED instead of NA when no operator exists.

Worth keeping as reference for a rewrite: the Subscription → `status.installedCSV` → deployments → pods traversal, and the priv/non-priv sample-operator fixture concept (redesigned to be self-contained for CI). A fresh implementation should live in the security category per the issue. Thanks @williscool for the groundwork and the research notes — they'll make the rewrite easier.

Issue #1597, opened 2022-08-11
[BUG] Reasonable Image Size -- Could Not Open Tar File


EricLo-417:
**Describe the bug**
Looks like the cnf-testsuite cannot open the tar version of the chart to check image size. It appears to store it in "tmp". It does mention image not found, however the pod is running with the specified image

**To Reproduce**

1. Run "reasonable_image_size"

**Expected behavior**
It would find the image size

**Logs**
```
I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec command: kubectl exec -n cnf-testsuite dockerd -t -- docker save localhost:32000/matrixx-activemq:5250-SNAPSHOT -o /tmp/image.tar
I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec stderr: Error response from daemon: reference does not exist
command terminated with exit code 1

I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec command: kubectl exec -n cnf-testsuite dockerd -t -- gzip -f /tmp/image.tar
I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec stderr: gzip: /tmp/image.tar: No such file or directory
command terminated with exit code 1

I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec command: kubectl exec -n cnf-testsuite dockerd -t -- wc -c /tmp/image.tar.gz | awk '{print$1}'
I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: KubectlClient.exec stderr: wc: /tmp/image.tar.gz: No such file or directory
command terminated with exit code 1

I, [2022-05-19 01:33:28 -04:00 #2783727]  INFO -- cnf-testsuite: compressed_size: localhost:32000/matrixx-activemq:5250-SNAPSHOT = ''
E, [2022-05-19 01:33:28 -04:00 #2783727] ERROR -- cnf-testsuite: [31minvalid compressed_size: localhost:32000/matrixx-activemq:5250-SNAPSHOT = '', Invalid Int64: [0m
[...]

PR #1800, opened 2023-06-20
[bug/1597] Allow configuring insecure docker registries and image registry FQDNs

HashNuke:
## Issues
Refs: #1597

## Description

The `reasonable_image_size` test would throw an `Int64` error. This was due to two issues:

#### 1. Insecure access was not configurable for the docker daemon being setup by the testsuite

Private registries hosted on the cluster usually have to be trusted to use registry API endpoints via HTTP. The docker daemon has to be configured to be allowed to access the private image registry on the cluster via HTTP endpoints.

#### 2. Docker daemon requires FQDN to use a private registry

The Docker daemon is setup by the testsuite on the `cnf-testsuite` namespace. For the `reasonable_image_size` test, when the testsuite is referring to an image registry on another namespace, the testsuite requires the service discovery url for the service. It cannot access the registry with just the service name.

So if the helm chart mentions `foobar:5000/coredns:1.6.7`, the docker daemon wouldn't know which namespace the service is on.

## Solution in this PR

This PR adds support for two options in `cnf-testsuite.yml` to support private image registries.

### `image_registry_fqdns` option

When using a private registry hosted on the cluster, the image references in the CNF's helm chart may refer to the registry host with the Kubernetes service name alone. The CNF Testsuite runs a docker daemon in a separate `cnf-testsuite` namespace on the Kubernetes cluster. So it is required for the testsuite to be aware of the FQDN of the service, along with the port.

Use this option to configure FQDNs of image registries for the testsuite to access.

If the CNF's helm charts use the image url `foobar:5000/hello:latest`, then please use the configuration in below in `cnf-testsuite.yml` to provide the FQDN mapping for the image registry.

```yaml
image_registry_fqdns:
    "foobar:5000": "foobar.default.svc.cluster.local:5000"
```

> *The above example assumes that the `foobar` registry service is running on the `default` namespace.*

### `docker_insecure_registries` option

The docker client expects the image registries to be using an HTTPS API endpoint. This option is used to configure insecure registries that the docker client should be allowed to access.

Please use this option to configure Docker to use HTTP to access the registry API.

For an image registry service named `foobar`, running in `default` namespace, on port `5000`, the following would be the expected configuration.

```yaml
docker_insecure_registries: ["foobar.default.svc.cluster.local:5000"]
```

## Validating the issue & verifying this PR

The git repo <https://git.sr.ht/~akash/cnf-testsuite-issue-1597> has notes on reproducing this issue.

### Example CNF to reproduce the error

```shell
./cnf-testsuite cnf_setup cnf-path=sample-coredns
./cnf-testsuite reasonable_image_size
# This should fail
```

### Example CNF to test the new `cnf-testsuite.yml` options that resolves the error

There is a second CNF in the same git repo mentioned above which has a config that uses new testsuite options introduced in this PR.

```shell
./cnf-testsuite cnf_setup cnf-path=sample-coredns-FIXED
./cnf-testsuite reasonable_image_size
# This should pass
```

## Checklist

 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [x] Documentation update

## Checklist:
**Documentation**
- [x] I have updated the documentation accordingly.
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -153,6 +153,7 @@ jobs:
         DOCKERHUB_PASSWORDS: ${{ secrets.DOCKERHUB_PASSWORDS }}
         DOCKERHUB_EMAIL: ${{ secrets.DOCKERHUB_EMAIL }}
         IMAGE_REPO: ${{ secrets.IMAGE_REPO }}
+        CNF_TESTSUITE_DIR: "/tmp/cnf-testsuite-home"
       run: |
         USERNAME_ARRAY=($DOCKERHUB_USERNAMES)
         PASSWORD_ARRAY=($DOCKERHUB_PASSWORDS)
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -153,7 +153,6 @@ jobs:
         DOCKERHUB_PASSWORDS: ${{ secrets.DOCKERHUB_PASSWORDS }}
         DOCKERHUB_EMAIL: ${{ secrets.DOCKERHUB_EMAIL }}
         IMAGE_REPO: ${{ secrets.IMAGE_REPO }}
-        CNF_TESTSUITE_DIR: "/tmp/cnf-testsuite-home"
       run: |
         USERNAME_ARRAY=($DOCKERHUB_USERNAMES)
         PASSWORD_ARRAY=($DOCKERHUB_PASSWORDS)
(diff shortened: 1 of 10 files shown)


agentpoyo (approved):
LGTM, will test before final merge.

Issue #1735, opened 2023-02-07
[BUG] When increase fails, the testsuite should not attempt a decrease during the `increase_decrease_capacity` test


agentpoyo:
**Describe the bug**
It appears if the increase fails when running the `increase_decrease_capacity` test, it still attempts to run the decrease test. The testsuite should skip this as the timeout can cause the testsuite from running more efficiently.

**To Reproduce**
Any CNF that fails to increase the capacity, you will see that it attempts to decrease back to 1.

**Expected behavior**
The testsuite should skip decrease if the increase was unsuccessful.

Snippet from debug showing the kubectl command to scale to 3, once it fails, it then attempts to scale back to 1 replica:

```
I, [2023-02-07 17:18:44 +00:00 #39324]  INFO -- cnf-testsuite-change_capacity:resource: Deployment/coredns-coredns; namespace: cnfspace
I, [2023-02-07 17:18:44 +00:00 #39324]  INFO -- cnf-testsuite-change_capacity:capacity: Base replicas: 3; Target replicas: 1
I, [2023-02-07 17:18:44 +00:00 #39324]  INFO -- cnf-testsuite: KubectlClient::Scale.command command: kubectl scale Deployment.v1.apps/coredns-coredns --replicas=3 -n cnfspace
D, [2023-02-07 17:18:44 +00:00 #39324] DEBUG -- cnf-testsuite: KubectlClient::Scale.command output: 
I, [2023-02-07 17:18:44 +00:00 #39324]  INFO -- cnf-testsuite: KubectlClient::Scale.command stderr: E0207 17:18:44.272219   40670 memcache.go:238] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused

I, [2023-02-07 17:19:17 +00:00 #39324]  INFO -- cnf-testsuite: KubectlClient::Scale.command command: kubectl scale Deployment.v1.apps/coredns-coredns --replicas=1 -n cnfspace
D, [2023-02-07 17:19:17 +00:00 #39324] DEBUG -- cnf-testsuite: KubectlClient::Scale.command output: 
I, [2023-02-07 17:19:17 +00:00 #39324]  INFO -- cnf-testsuite: KubectlClient::Scale.command stderr: E0207 17:19:17.856503   41318 memcache.go:238] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[...]

PR #1862, opened 2024-01-24
increase_decrease_capacity: Skip decrease on unsuccessful increase

kosstennbl:
If capacity increase have failed - capacity decrease test should be skipped.
Add check for capacity increase before capacity decrease test. Redisign logging and error handling at the end of the test. Increase default wait_count_value to 45s
(was 30, which is too small)
Fixes: #1735 

Tested by running increase_decrease_capacity test on coredns cnf, both successfully and unsuccessfully (with lower wait time it fails)

## Issues:
Refs: #1735

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/workload/compatibility.cr
+++ b/src/tasks/workload/compatibility.cr
@@ -193,10 +193,14 @@ task "increase_decrease_capacity" do |t, args|
     testsuite_task = "increase_decrease_capacity"
     Log.for(testsuite_task).info { "Starting test" }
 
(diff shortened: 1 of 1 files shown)


HashNuke (approved):
LGTM ~!

Thank you for different & clear messages for different scenarios when increase/decrease fails.

Issue #1922, opened 2024-03-06
[Feature] Uniform way of logging spec test output


macaktom:
**Is your feature request related to a problem? Please describe.**
Right now, there is a lot of repetition regarding spec test outputs and there are multiple ways of logging outputs (Log and LOGGING).

Good example could be cluster_api_spec.cr

![image](https://github.com/cnti-testcatalog/testsuite/assets/43893713/56c1b0a9-f7d7-481e-a154-341876d07c04)

1) Log.info and LOGGING.info -> multiple ways how to log result
2) multiple repetition of assigning result to variable (resp, response_s) and it's logging.

**Describe the solution you'd like**

1) Think of uniform way to output spec results -> for example wrapper function, which would encapsulate result of shell command and output it if needed (wrapper would encapsulate variable assignment and output of that variable).
2) Replace instances with Log and Logging with implemented wrapper function.

**How will this be tested? aka Acceptance Criteria (optional)**
1) Running all spec tests in debug mode
2) Running all spec tests in error mode

(optional: unnecessary for things like some version bumps etc.)

Goal is to avoid repetition and have one way to log output of shell commands. Maybe this could also be applicable to workload and platform tests (need to be checked).

horecoli:
In some logs from spec tests I saw this (just snippet from middle of the log):
`I, [2024-02-28 13:51:42 +01:00 #765697]  INFO -- cnf-testsuite: task_type x: bonus acc: cert
I, [2024-02-28 13:51:42 +01:00 #765697]  INFO -- cnf-testsuite: task_type: bonus
I, [2024-02-28 13:51:42 +01:00 #765697]  INFO -- cnf-testsuite: upsert_task: task: no_local_volume_configuration has status: passed and is awarded: 1 points. Runtime: 276 seconds
I, [2024-02-28 13:51:42 +01:00 #765697]  INFO -- cnf-testsuite: results yaml: {"name" => "cnf testsuite", "testsuite_version" => "privileged_test_1906-2024-02-28-135141-e54218b7", "status" => nil, "command" => "/home/horecoli/git/cnf-testsuite/cnf-testsuite no_local_volume_configuration verbose", "points" => nil, "exit_code" => 0, "items" => [{"name" => "no_local_volume_configuration", "status" => "passed", "type" => "bonus", "points" => 1}]}

[...]

Issue #1495, opened 2022-06-08
[BUG] Remove Log.info wrappers when running shell commands in spec tests


HashNuke:
## Describe the bug
Running shell commands in `Log.info { }` would result in the command being skipped when the log level is higher or not set.
So when the log level is higher, then the spec test would fail.

<img width="1305" alt="CleanShot 2022-06-08 at 18 38 34@2x" src="https://user-images.githubusercontent.com/84005/172624339-bc5b4efa-1c99-4503-be94-2d380574eb63.png">

macaktom:
I can look into this task.

PR #1923, opened 2024-03-06
Spec tests shell command fix

macaktom:
## Description
This PR will remove shell commands from Log.info wrapper.

Part of PR is also removal of command `./cnf-testsuite setup`, which is used two times in microservice_spec.cr **before_all** construct.

PR only provides quick fix of issue, where spec tests fail, if shell command is run in Log.info and log level error is used. To avoid repetition and add uniform way of logging spec tests I created different task for that (#1922), as that's not point of this fix.

## Issues:
Refs: #1495

## How has this been tested:
1) Run spec tests in debug mode
2) Run spec tests in error mode

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


HashNuke (changes requested):
When using backticks to run shell commands the output is printed to stdout.

So when log level is set to info, the output would be printed as log lines (to stdout), as well as to stdout again (for the backticks).

For the intended outcome of this PR, we could use the Process module.

There is also a ShellCmd module within the project that can be used/repurposed for the intended outcome of this PR. This wraps the Process module’s function call.

HashNuke:
I was wrong about the behaviour of backticks. I just tested it. But it still applies to stderr. If we use backticks, we will get some lines in stderr and some lines in stdout (in the log block) if the command outputs both.

This is the `ShellCmd` helper module that runs the command, and then outputs the command's output as a part of Log blocks - https://github.com/cnti-testcatalog/testsuite/blob/83f08b67171cfeeac84e0e4d2af60ff521ef1393/src/tasks/utils/utils.cr#L15-L36

If there is missing functionality with the ShellCmd module, please feel free to update it as required.

CleanShot 2024-04-02 at 20 59 46@2x

kosstennbl:
New PR: https://github.com/cnti-testcatalog/testsuite/pull/1954

Issue #1945, opened 2024-04-03
[Feature] Documentation rework


kosstennbl:
Proposal is to join all files that contain test-related documentation into one file with fixed formatting that will contain all info for developers and users.
Currently, test-related documentation is located in these files:  LIST_OF_TESTS.md, RATIONALE.md, USAGE.md, TEST-CATEGORIES.md.

### Issues with current documentation:
- Hard to navigate (to get complete information about the test you need to browse multiple files)
- Redundant information (links to other files, repetition of test details)
- Hard to maintain (changes for one test result in need for change in multiple documentation files)
- Hard to verify (some or all info about test might be missing, it's time-consuming to validate that all documentation is present)

### Proposals:
- Join documentation into TESTSUITE-TESTS.md file with fixed structure and formatting.
- Create GitHub pipeline that will validate that file and ensure that structure isn't broken and all needed sections are present

In addition to fixing current issues, new format for documentation will have these possibilities in future (not in the scope of the issue):
- Usage of documentation in code, directly parsed from the documentation. For example "remediation" in case if a test fails.
- Moving content of points.yml, including tags and points, to the TESTSUITE-TESTS.md, and using this in the code
- Possible automated generation of specialized documents, that will contain only subset of information for tests.

About having separate file with reasoning for users: is it really needed? If documentation is nicely structured, then it would be easy for user to find all desired information including reasoning for the tests.

Proposed structure of the documentation:

# Category 1
Some info about the category
#### Overview:
(task_name) Description of the test, what does it do
#### Reasoning:
Reasoning for the tests, why is it needed, how does it help for cloud-nativeness
#### Remediation:
What needs to be done if the test is failing.
#### Usage:
Command for running the test, extra dependencies if present.
## Test 2
...

kosstennbl:
WIP: example of the file https://github.com/kosstennbl/cnf-testsuite/blob/test-doc-unification/docs/TEST_DOCUMENTATION.md

PR #1955, opened 2024-04-08
docs: Unification of documentation

kosstennbl:
Join all test-related documentation into one file to unified format.
Format is mentioned in related issue and is open for discussion.

## Issues:
Refs: #1945

Does not afflict code, needs no testing.


taylor:
What about the old docs? Duplication

kosstennbl:
> What about the old docs? Duplication

We can move them to some folder "old_docs" in case they would be needed

agentpoyo (changes requested):
There needs to be horiontal lines or some type of dividers between tests in the documentation. The flow of the document is hard to follow knowing when a test begins and ends in the docs since every section has Usage, Remediation, Reasoning. 

Perhaps some of the redundancy could be hidden with drop downs as well for each test to limit the amount of scrolling within the doc as well.

Also, "reasoning" sections should be labeled as "Rationale" perhaps? There's a reason we used the term Rationale, reasoning just sounds like another way to say Description of the test. Thoughts @wavell @denverwilliams @HashNuke @taylor

agentpoyo (changes requested):
The table of contents is very long. Perhaps limit the amount listed down to the categories instead of every single test. See https://github.com/cnti-testcatalog/testsuite/blob/main/USAGE.md#compatibility-installability-and-upgradability-tests TOC. 

Also, there seems to be some valid information missing from USAGE.md that explains the output from tests, etc that I feel is still good information in the docs: https://github.com/cnti-testcatalog/testsuite/blob/main/USAGE.md#syntax-for-running-any-of-the-tests and the next output should get carried over into the unified doc somewhere. 

Anything in the other docs that are valid should be carried over as well.

martin-mat:
> > What about the old docs? Duplication
> 
> We can move them to some folder "old_docs" in case they would be needed

I'm voting for removing old docs. This would introduce unnecessary redundancy and mess. Whoever needs to look into old docs can do it in git history/older tags.

taylor:
I am down for removing duplicate docs.  I think the missing/removed information from RATIONALE and other places should be added to this combined document before it is merged though.

kosstennbl:
> Also, there seems to be some valid information missing from USAGE.md that explains the output from tests, etc that I feel is still good information in the docs: https://github.com/cnti-testcatalog/testsuite/blob/main/USAGE.md#syntax-for-running-any-of-the-tests and the next output should get carried over into the unified doc somewhere.

About USAGE.md: my intention wasn't to remove USAGE.md completely. Idea is to remove information that covers test categories and tests, but leave all other info in place.

Agree with deleting RATIONALE.md LIST_OF_TESTS.md and TEST-CATEGORIES.md.

About missing information: two types of information weren't transferred as i consider them duplicate to other info in test:
* In RATIONALE.md, there is some text in *cursive* before test name, but it's shortened version of overview.
* In LIST_OF_TESTS.md, tests have expectation. I've not transferred that due to the information being mostly duplicate to reasoning and remediation. (But i see how it can still be useful, can be transferred as well if you consider it needed).

Reasoning vs rationale: to me - these terms are very similar in meaning, but if using "rationale" makes more sense to you - i have nothing against that.

 Agree with dividers, will add them. Dropdowns could be nice, but they would add significant amount of HTML tags to MD document, which doesn't look that good (in code) and could be harder to maintain:
```
<details>
<summary>How do I dropdown?</summary>
<br>
This is how you dropdown.
</details>
```
If we want header inside summary - it would be:

```
<summary><h3>How do I dropdown?</h3></summary>
```

kosstennbl:
@taylor @wavell @agentpoyo Please check if information that is missing is really needed. 

>About missing information: two types of information weren't transferred as i consider them duplicate to other info in test:
>
>  * In RATIONALE.md, there is some text in cursive before test name, but it's shortened version of overview.
>  * In LIST_OF_TESTS.md, tests have expectation. I've not transferred that due to the information being mostly duplicate to 
       reasoning and remediation. (But i see how it can still be useful, can be transferred as well if you consider it needed).

kosstennbl:
Missing info moved, old docs deleted.
Please rewiew, @HashNuke @taylor @wavell  @agentpoyo

.gitignore
```diff
@@ -1,4 +1,3 @@
-/docs/
```
agentpoyo, line 1: Can we remove this from the commits? We don't trigger github actions when MD files are updated but this change causes unecessary github action triggers.

agentpoyo (changes requested):

.gitignore
kosstennbl: Done

agentpoyo (changes requested):
@kosstennbl 

There are links to USAGE.md and likely other links scattered around the MD files not being removed that have not been updated. Merging this will break URL links to the files that are now being removed. These need to be all fixed before this is merged.

kosstennbl:
> There are links to USAGE.md and likely other links scattered around the MD files not being removed that have not been updated. Merging this will break URL links to the files that are now being removed. These need to be all fixed before this is merged.

Fixed links that are affected by the change, fixed USAGE.md TOC and added link from USAGE to TEST_DOCUMENTATION, fixed some of the links that pointed to the old cncf repo (don't see too much reason to create separate PR for that small change)

kosstennbl:
Resolved conflicts, ready for review.

martin-mat (changes requested):
TEST_DOCUMENTATION.md:
- dashboard-exposed dead link (the test should be removed)
- "Environment variables for timeouts:" missing in USAGE.md

other than that OK with me.

kosstennbl:
Nice catch, TOC is a bit hard to check (linter for that comes as the next PR)
ENV timeouts - was missing rebase. Both are fixed now.

martin-mat (approved):
lgtm

horecoli (approved):
lgtm

rich-l (approved):
lgtm

Issue #2022, opened 2024-05-11
Node drain test not starting due to unable to get chaos resources (ChaosExperiment.litmuschaos.io "node-drain" not found)


sysarch-repo:
Node drain test not starting due to chaos resources not found

**Steps to reproduce**
$ cnf-testsuite version
CNF TestSuite version: v1.2.0

$ cnf-testsuite node_drain
🎬 Testing: [node_drain]
< not progressing>

$ kubectl get nodes
```
NAME                           STATUS                     ROLES    AGE   VERSION
ip-10-0-116-213.ec2.internal   Ready                      <none>   39m   v1.28.8-eks-ae9a62a
ip-10-0-76-157.ec2.internal    Ready,SchedulingDisabled   <none>   39m   v1.28.8-eks-ae9a62a
```

$ kubectl describe chaosengine -n cnti dns-dserver-1c6b0d96
```
Name:         dns-dserver-1c6b0d96
Namespace:    cnti
Labels:       <none>
Annotations:  <none>
API Version:  litmuschaos.io/v1alpha1
Kind:         ChaosEngine
Metadata:
  Creation Timestamp:  2024-05-11T19:52:17Z
  Finalizers:
    chaosengine.litmuschaos.io/finalizer
  Generation:        2
  Resource Version:  4544
  UID:               e4bd93c5-290f-48f2-ae0f-d968f55b0590
Spec:
  Appinfo:
    Appkind:              deployment
    Applabel:             app.nti/pod-group=dns-dserver
    Appns:                cnti
  Chaos Service Account:  node-drain-sa
  Components:
    Runner:
      Resources:
  Engine State:  active
  Experiments:
    Name:  node-drain
    Spec:
      Components:
        Env:
          Name:   TOTAL_CHAOS_DURATION
          Value:  90
          Name:   TARGET_NODE
          Value:  ip-10-0-76-157.ec2.internal
        Resources:
        Status Check Timeouts:
  Job Clean Up Policy:  delete
Status:
  Engine Status:  initialized
  Experiments:    <nil>
Events:
  Type     Reason                         Age                  From            Message
  ----     ------                         ----                 ----            -------
  Normal   ChaosEngineInitialized         23m                  chaos-operator  Identifying app under test & launching dns-dserver-1c6b0d96-runner
  Warning  ChaosResourcesOperationFailed  116s (x19 over 23m)  chaos-operator  (chaos start) Unable to get chaos resources
```

Chaos operator logs:
```
[...]

PR #2024, opened 2024-05-13
node-drain: Fix link to experiment

kosstennbl:
In 69dedb30dd686ab2d1edffb4178ddb3f2c94e7e4, Litmus version was updated And all links for experiments were changed accordingly except for the node_drain. This commit fixes that.

ref: #2022

### DIFF
--- a/src/tasks/workload/state.cr
+++ b/src/tasks/workload/state.cr
@@ -307,12 +307,8 @@ task "node_drain", ["install_litmus"] do |t, args|
             KubectlClient::Apply.file("#{OFFLINE_MANIFESTS_PATH}/node-drain-experiment.yaml")
             KubectlClient::Apply.file("#{OFFLINE_MANIFESTS_PATH}/node-drain-rbac.yaml")
           else
-            # experiment_url = "https://hub.litmuschaos.io/api/chaos/#{LitmusManager::Version}?file=charts/generic/node-drain/experiment.yaml"
-            # https://raw.githubusercontent.com/litmuschaos/chaos-charts/v2.14.x/charts/generic/pod-network-latency/experiment.yaml
-            # https://raw.githubusercontent.com/litmuschaos/chaos-charts/v2.14.x/charts/generic/node-drain/experiment.yaml 
-            experiment_url = "https://raw.githubusercontent.com/litmuschaos/chaos-charts/#{LitmusManager::Version}/charts/generic/node-drain/experiment.yaml"
-            # rbac_url = "https://hub.litmuschaos.io/api/chaos/#{LitmusManager::Version}?file=charts/generic/node-drain/rbac.yaml"
-            rbac_url = "https://raw.githubusercontent.com/litmuschaos/chaos-charts/#{LitmusManager::Version}/charts/generic/node-drain/rbac.yaml"
+            experiment_url = "https://raw.githubusercontent.com/litmuschaos/chaos-charts/#{LitmusManager::Version}/faults/kubernetes/node-drain/fault.yaml"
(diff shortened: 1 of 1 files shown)


taylor:
Lgtm

martin-mat:
lgtm

martin-mat (approved):

taylor (changes requested):
The spec test before this change should fail and the spec should pass after this fix. 

If not the spec should be updated to produce these results before this is merged to ensure we have coverage.

martin-mat:
there is a good spec test for node_drain
https://github.com/cnti-testcatalog/testsuite/blob/main/spec/workload/resilience/node_drain_spec.cr
which behaves correctly (verified).

The real reason why the issue was not detected earlier during github actions is that they use one-node kind setup for testing. node_drain test needs multi-node setup and the spec tests "passes" because the test is "skipped".

⏭️  🏆SKIPPED: [node_drain] node_drain chaos test requires the cluster to have atleast two schedulable nodes 🗡️💀♻

The spec tests is happy with such skipping:
```
      if KubectlClient::Get.schedulable_nodes_list.size > 1
        (/(PASSED).*(node_drain chaos test passed)/ =~ result[:output]).should_not be_nil
      else
        (/(SKIPPED).*(node_drain chaos test requires the cluster to have atleast two)/ =~ result[:output]).should_not be_nil
      end
```

So I propose to adapt github actions so they run on kind with 2 schedulable nodes
Since it is more generic adaptation I suggest to handle this in a separate ticket.
#2026

daniel-wilmes:
@martin-mat I have verified that the fix for node drain works as a single test and in the cert command. However, I will say the cert command does not ever finish, which appears to be a seperate issue from node drain. It may pertain to either sig_term_handled, zombie_handled, or specialized_init_system. The logging doesn't appear to indicate where we are stuck. But for this ticket I think the fix for node_drain should go in.

`---
name: cnf testsuite
testsuite_version: node-drain-fix-2024-05-15-142132-3258c691
status: 
command: /home/dwilmes/.mtx/konstruxx/working/tests/testCHF/cnf-testsuite cert essential
points: 100
exit_code: 0
items:
- name: increase_decrease_capacity
  status: passed
  type: essential
  points: 100
- name: volume_hostpath_not_found
  status: passed
  type: essential
  points: 100
- name: node_drain
  status: passed
  type: essential
  points: 100
- name: privileged_containers
  status: passed
  type: essential
  points: 100
- name: non_root_containers
  status: failed
  type: essential
  points: 0
- name: cpu_limits
  status: passed
  type: essential
  points: 100
- name: memory_limits
  status: passed
  type: essential
  points: 100
- name: hostpath_mounts
  status: passed
  type: essential
  points: 100
- name: container_sock_mounts
  status: passed
  type: essential
  points: 100
- name: selinux_options
  status: na
  type: essential
  points: 0
- name: hostport_not_used
  status: passed
  type: essential
  points: 100
- name: hardcoded_ip_addresses_in_k8s_runtime_configuration
  status: passed
  type: essential
  points: 100
- name: latest_tag
  status: passed
  type: essential
  points: 100
- name: log_output
  status: passed
  type: essential
  points: 100
- name: specialized_init_system
  status: failed
  type: essential
  points: 0
`

taylor:
@daniel-wilmes please open a new issue for
> However, I will say the cert command does not ever finish, which appears to be a seperate issue from node drain. It may pertain to either sig_term_handled, zombie_handled, or specialized_init_system. The logging doesn't appear to indicate where we are stuck. 

You can try disabling those with ~testname (eg. ~sig_term_handled) and isolate which test needs to be investigated. Add that info to the new ticket you create.

cc: @Smitholi67

taylor:
@martin-mat good catch on the kind in github actions. https://github.com/cnti-testcatalog/testsuite/pull/2024#issuecomment-2112309884

taylor (approved):
lgtm and specs pass

Issue #2036, opened 2024-05-17
[Feature] Refactor sample_setup


kosstennbl:
**Is your feature request related to a problem? Please describe.**
In cnf_manager.cr, sample_setup function is too long and complex (338 LoC). It's hard to comprehend and maintain.

**Describe the solution you'd like**
Refactoring of the sample_setup function, extracting code to multiple smaller functions.

kosstennbl:
If serious improvements for setup process would be planned - i consider that these problems should be dealt with first:

1. Airgap mode is still present, and considering the fact that it isn't used anymore - it should be removed.
   - image_pull method has an unused parameter connected to airgap mode (offline is always a string which evaluates to true) 
2. For some reason - we have manifest installation parameters and methods in Helm repository. IMO - they should be extracted and separated.
3. Code for HelmDirectory installation method should reuse code for HelmChart installation method
   - Why even have them as different methods? IMO - they should be joined into "Helm" installation method, and the app will try to parse whatever is given (if Chart.yml address - nice, if dir - search for chart, if tar - untar and search).
4. Jaeger and ORAN checks should be extracted from cnf setup. If current way is unsuitable to be used outside of setup then different way of tracking these parameters should be found. These parameters are needed only for specific CNFs and have nothing to do with the setup process.

PR #2050, opened 2024-05-22
cnf_manager: Refactor sample_setup

kosstennbl:
## Description
Change instalation time measurement method to reduce complexity Extract waiting for all resources and liveness_time calculation Extract elapsed time config map creation
Add comments as markers for different stages of sample setup

Markers are to be removed once method is simplified enough.

## Issues:
Refs: #2036

## How has this been tested:
Successfully built

## Types of changes:
Refactor

## Checklist:
**Documentation**
No documentation updates required.

### DIFF
--- a/src/tasks/utils/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager.cr
@@ -775,6 +775,9 @@ module CNFManager
 
   #sample_setup({config_file: cnf_path, wait_count: wait_count})
   def self.sample_setup(cli_args)
+
+    # PREPARATION FOR INSTALL
+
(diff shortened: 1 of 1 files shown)


kosstennbl:
In my opinion - changes are simple enough to be coupled together like this.
Further changes will come as this will be merged.

kosstennbl:
Added comment to the referenced issue.
I consider that changes in this PR are useful and can be considered as "low-hanging fruit" in topic of refactoring the setup process.
For full refactor - multiple project-wide problems (check #2036) should be solved first, only then it seems sensible to me to fix this function.

Issue #1947, opened 2024-04-04
[BUG] Erroneous file globbing in cnf_manager - export_published_chart


svteb:
**Describe the bug**
While running the `database persistence` spec test I ran into a rather unexpected bug with the `TarClient module`. The `CNFManager.export_published_chart` function uses this construct:

```
def self.export_published_chart(config, cli_args)
...
   tgz_name = "#{Helm.chart_name(helm_chart)}-*.tgz"
...
```

This causes the tar module to fail in case multiple versions of a helm chart archive are present in the `testsuite` directory.

To best describe the issue, let me provide you with an example. suppose the mysql sample cnf is going to be deployed, the `TarClient module` executes the following command to export the chart to the necessary folder in `/cnfs/mysql/exported_chart`:

```
tar -xvf mysql-*.tgz -C /cnfs/mysql/exported_chart ~>
tar -xvf mysql-9.0.1.tgz mysql-10.1.1.tgz mysql-10.1.2.tgz ... -C /cnfs/mysql/exported_chart
```

Supposing there are multiple mysql-*.tgz helm charts in the `testsuite` directory the tar command fails.

**To Reproduce**

I believe this issue might affect other spec tests/maybe even the main program, but I've only noticed it with the `database_persistence` spec test:

0. `export LOG_LEVEL=debug`
1. `cd testsuite`
2. `touch mysql-x.y.z.tgz` or `cp` the actual mysql helm chart and rename it
3. `crystal spec spec/workload/state_spec.cr:43 | tee results/database_persistence.txt # database persistence spec test`
4. During the program runtime you should be capable of seeing that the `/cnfs/mysql/exported_chart` directory is not populated with files from the helm chart. The fact that the mysql statefulset successfully deploys is not indicative of this not being an error.
5. The test fails, upon observing the `results/database_persistence.txt` file you may see log outputs like this:

```
I, [2024-04-04 08:10:48 +00:00 #1788708]  INFO -- cnf-testsuite: TarClient.untar current directory: /home/ubuntu/testsuite
I, [2024-04-04 08:10:48 +00:00 #1788708]  INFO -- cnf-testsuite: TarClient.untar command: tar  -xvf mysql-*.tgz -C /home/ubuntu/testsuite/cnfs/mysql/exported_chart
I, [2024-04-04 08:10:48 +00:00 #1788708]  INFO -- cnf-testsuite: TarClient.untar output:
[...]

PR #2052, opened 2024-05-23
Fix: export_published_chart correctly detects and exports tar file

svteb:
## Description
Removed file globbing as it caused errors and added a new function that will detect a specific tgz file to be untared. Removed the line: `FileUtils.rm_rf(tgz_name)` as it did not work previously and now that globbing has been fixed it would cause undesired behavior (honestly I was not able to figure out why it is there at all).

## Issues:
Refs: #1947 

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [x] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
    * [x] Minikube cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**
- [x] Tasks in issue are checked off

### DIFF
--- a/src/tasks/utils/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager.cr
@@ -728,15 +728,14 @@ module CNFManager
       tgz_name = tar_info[:tar_name]
     elsif output_file && !output_file.empty?
       config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), generate_tar_mode: true) 
-      tgz_name = "#{Helm.chart_name(helm_chart)}-*.tgz"
+      tgz_name = get_tgz_name(helm_chart)
     else
       config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file))
-      tgz_name = "#{Helm.chart_name(helm_chart)}-*.tgz"
+      tgz_name = get_tgz_name(helm_chart)
     end
     Log.info { "tgz_name: #{tgz_name}" }
 
     unless input_file && !input_file.empty?
-      FileUtils.rm_rf(tgz_name)
       helm_info = Helm.pull(helm_chart) 
       unless helm_info[:status].success?
         puts "Helm pull error".colorize(:red)
@@ -744,7 +743,7 @@ module CNFManager
       end
     end
 
-    TarClient.untar(tgz_name,  "#{destination_cnf_dir}/exported_chart")
+    TarClient.untar(tgz_name, "#{destination_cnf_dir}/exported_chart")
 
     Log.for("verbose").info { "mv #{destination_cnf_dir}/exported_chart/#{Helm.chart_name(helm_chart)}/* #{destination_cnf_dir}/exported_chart" } if verbose
     Log.for("verbose").debug {
@@ -1257,6 +1256,12 @@ module CNFManager
   def self.resources_includes?(resource_keys, kind, name, namespace)
     resource_key = "#{namespace},#{kind}/#{name}".downcase
     resource_keys.includes?(resource_key)
+  end
+
+  def self.get_tgz_name(helm_chart)
+    tar_files = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+    raise "No tar files found" if tar_files.empty?
+    tar_files.last
   end
 
   class HelmDirectoryMissingError < Exception
--- a/src/tasks/utils/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager.cr
@@ -724,10 +724,10 @@ module CNFManager
 
     # delete previous tgz files and pull the helm chart
     unless input_file && !input_file.empty?
-      files_to_delete = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+      files_to_delete = tgz_files(helm_chart)
       files_to_delete.each do |file|
         FileUtils.rm(file)
-        puts "Deleted: #{file}"
+        Log.info { "Deleted: #{file}" }
       end
 
       helm_info = Helm.pull(helm_chart) 
@@ -1267,15 +1267,20 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
-  def self.get_tgz_name(helm_chart)
-    tgz_files = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
-  
-    # In case no tgz file is present (should only happen if there was a prior helm failure).
-    if tgz_files.empty?
-      Log.error { "No .tgz files found for #{Helm.chart_name(helm_chart)}" }
-      raise TarFileNotFoundError.new(Helm.chart_name(helm_chart))
-    end
+  def self.tgz_files(helm_chart)
+    Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+  end
 
+  def self.get_tgz_name(helm_chart)
+    tgz_files = tgz_files(helm_chart)
+    
+    if tgz_files.empty?
+      Log.error { "No .tgz files found for #{Helm.chart_name(helm_chart)}-*.tgz" }
+      raise TarFileNotFoundError.new(Helm.chart_name(helm_chart))
+    elsif tgz_files.size > 1
+      Log.info { "Multiple .tgz files found: #{tgz_files.join(", ")}" }
+    end
+  
     tgz_files.first
   end
 
@@ -1289,7 +1294,7 @@ module CNFManager
 
   class TarFileNotFoundError < Exception
     def initialize(chart_name)
-      super("No .tgz files found for chart #{chart_name}")
+      super("No .tgz files found for chart #{chart_name}-*.tgz")
     end
   end
 end
--- a/src/tasks/utils/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager.cr
@@ -721,6 +721,23 @@ module CNFManager
 
     input_file = cli_args[:input_file]
     output_file = cli_args[:output_file]
+
+    # delete previous tgz files and pull the helm chart
+    unless input_file && !input_file.empty?
+      files_to_delete = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+      files_to_delete.each do |file|
+        FileUtils.rm(file)
+        puts "Deleted: #{file}"
+      end
+
+      helm_info = Helm.pull(helm_chart) 
+      unless helm_info[:status].success?
+        puts "Helm pull error".colorize(:red)
+        raise "Helm pull error"
+      end
+    end
+
+    # get the tgz file name
     if input_file && !input_file.empty?
       # todo add generate and set tar as well
       config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), airgapped: true) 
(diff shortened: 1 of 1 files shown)


src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1258,12 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
+  def self.get_tgz_name(helm_chart)
+    tar_files = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+    raise "No tar files found" if tar_files.empty?
```
martin-mat, line 1263: more details about what filename was not found would make future troubleshooting easier.

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1258,12 @@ module CNFManager
...
+    tar_files.last
```
martin-mat, line 1264: A warning in case that multiple files found?

src/tasks/utils/cnf_manager.cr
```diff
@@ -728,23 +728,22 @@ module CNFManager
...
+      tgz_name = get_tgz_name(helm_chart)
     else
       config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file))
-      tgz_name = "#{Helm.chart_name(helm_chart)}-*.tgz"
+      tgz_name = get_tgz_name(helm_chart)
     end
     Log.info { "tgz_name: #{tgz_name}" }
 
     unless input_file && !input_file.empty?
-      FileUtils.rm_rf(tgz_name)
```
martin-mat, line 739: I guess the intention here was to remove an existing helm chart package to make sure it is pulled successfully?

martin-mat (changes requested):

src/tasks/utils/cnf_manager.cr
svteb: Resolved by adding a cleaner exception:

```
class TarFileNotFoundError < Exception
  def initialize(chart_name)
    super("No .tgz files found for chart #{chart_name}")
  end
end
```

src/tasks/utils/cnf_manager.cr
svteb: View the new change/read my new comment, this state is no longer possible.

src/tasks/utils/cnf_manager.cr
svteb: Yes, the intention was to remove all prior .tgz archives (did not work because this function is incompatible with file globbing). I backtracked on this removal and remade it so it would work.

svteb:
Comments from @martin-mat were correct in pointing out some unhandled cases/lack of information provided. I've come to understand the intentions behind `FileUtils.rm_rf(tgz_name)` and decided to bring it back:

```
unless input_file && !input_file.empty?
      files_to_delete = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz") ~> addition
      files_to_delete.each do |file|
        FileUtils.rm(file)
        puts "Deleted: #{file}"
      end ~> end of addition

      helm_info = Helm.pull(helm_chart) 
      unless helm_info[:status].success?
        puts "Helm pull error".colorize(:red)
        raise "Helm pull error"
      end
end
```

It was necessary to do it like this because `FileUtils.rm_rf` does not work with file globbing which was originally present. This addition also removed the necessity of checking for state where multiple versions of some helm archive were present (as they get deleted and the newest one gets pulled subsequentially).

Another mistake from the previous commit, that was undiscovered, was the necessity of moving this entire block before the block where tgz archive is being discovered, as the `Helm.pull(helm_chart)` command is what downloads the tgz archive. This also makes the sequence of steps somewhat more logical:

`helm_pull -> discover_tgz -> extract_tgz` whereas the previous sequence was `discover_tgz -> helm_pull -> extract_tgz`.


Furthermore I added an exception that prints better information in case no tgz file is discovered.

src/tasks/utils/cnf_manager.cr
martin-mat: I checked the updated code but the point with possible multiple files is still here. At least issue a warning.

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,18 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
+  def self.get_tgz_name(helm_chart)
+    tgz_files = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
+  
+    # In case no tgz file is present (should only happen if there was a prior helm failure).
+    if tgz_files.empty?
+      Log.error { "No .tgz files found for #{Helm.chart_name(helm_chart)}" }
+      raise TarFileNotFoundError.new(Helm.chart_name(helm_chart))
```
martin-mat, line 1276: Better report as much information as we have now, like:
No files with glob "#{Helm.chart_name(helm_chart)}-*.tgz" found

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,18 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
+  def self.get_tgz_name(helm_chart)
+    tgz_files = Dir.glob("#{Helm.chart_name(helm_chart)}-*.tgz")
```
martin-mat, line 1271: That glob pattern is used already twice so it's worth to have it on one place (a new helper method/getter?)

martin-mat (changes requested):

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,23 @@ module CNFManager
...
+  end
+
+  def self.get_tgz_name(helm_chart)
+    tgz_files = tgz_files(helm_chart)
+    
+    if tgz_files.empty?
+      Log.error { "No .tgz files found for #{Helm.chart_name(helm_chart)}-*.tgz" }
+      raise TarFileNotFoundError.new(Helm.chart_name(helm_chart))
+    elsif tgz_files.size > 1
+      Log.info { "Multiple .tgz files found: #{tgz_files.join(", ")}" }
```
martin-mat, line 1281: Log.warning
 plus note which one is being taken
?

src/tasks/utils/cnf_manager.cr
martin-mat: comment still valid

martin-mat (changes requested):

src/tasks/utils/cnf_manager.cr
svteb: Later on in the code the picked tgz file is logged.

src/tasks/utils/cnf_manager.cr
```diff
@@ -721,30 +720,39 @@ module CNFManager
...
+      end
+    end
+
+    # get the tgz file name
     if input_file && !input_file.empty?
       # todo add generate and set tar as well
       config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), airgapped: true) 
       tar_info = AirGap.tar_info_by_config_src(helm_chart)
       tgz_name = tar_info[:tar_name]
     elsif output_file && !output_file.empty?
```
martin-mat, line 745: simplify:
```
else 
      config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), generate_tar_mode: output_file && !output_file.empty?) 
      tgz_name = get_tgz_name(helm_chart)
end
```

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,27 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
+  def self.tgz_files(helm_chart)
+    Dir.glob(helm_tgz_expression(helm_chart))
+  end
+
+  def self.get_tgz_name(helm_chart)
+    tgz_files = tgz_files(helm_chart)
```
martin-mat, line 1275: avoid same identifiers for different things

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,27 @@ module CNFManager
     resource_keys.includes?(resource_key)
   end
 
+  def self.tgz_files(helm_chart)
```
martin-mat, line 1270: be more explicit with method naming, for example
```
def self.find_tgz_files(helm_chart)
```

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,27 @@ module CNFManager
...
+    Dir.glob(helm_tgz_expression(helm_chart))
+  end
+
+  def self.get_tgz_name(helm_chart)
```
martin-mat, line 1274: be more descriptive/explicit with method naming, for example
```
def self.get_and_verify_tgz_name(helm_chart)
```

src/tasks/utils/cnf_manager.cr
```diff
@@ -1259,6 +1267,27 @@ module CNFManager
...
+      Log.error { "No .tgz files found for #{helm_tgz_expression(helm_chart)}" }
+      raise TarFileNotFoundError.new(Helm.chart_name(helm_chart))
+    elsif tgz_files.size > 1
+      Log.warn { "Multiple .tgz files found: #{tgz_files.join(", ")}" }
+    end
+  
+    tgz_files.first
+  end
+
+  def self.helm_tgz_expression(helm_chart)
```
martin-mat, line 1287: better
```
def self.get_helm_tgz_expression(helm_chart)
```

martin-mat (changes requested):

src/tasks/utils/cnf_manager.cr
svteb: Good observation.

src/tasks/utils/cnf_manager.cr
svteb: Resolved

src/tasks/utils/cnf_manager.cr
svteb: Renamed

src/tasks/utils/cnf_manager.cr
svteb: Renamed

src/tasks/utils/cnf_manager.cr
svteb: Renamed

martin-mat (approved):
lgtm, good refactoring!

src/tasks/utils/cnf_manager.cr
```diff
@@ -721,30 +720,36 @@ module CNFManager
...
+    # get the tgz file name
+    if input_file && !input_file.empty?
+      # todo add generate and set tar as well
+      config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), airgapped: true) 
+      tar_info = AirGap.tar_info_by_config_src(helm_chart)
+      tgz_name = tar_info[:tar_name]
+    else
+      config = CNFManager::Config.parse_config_yml(CNFManager.ensure_cnf_testsuite_yml_path(config_file), generate_tar_mode: output_file && !output_file.empty?) 
+      tgz_name = get_and_verify_tgz_name(helm_chart)
+    end
```
kosstennbl, line 748: I think that these "if" statements should be joined and each logical branch marked with comments:
```
if input_file && !input_file.empty?   # Input file present, Tar installation mode
    ...
else   # Input file absent, Pulled chart installation mode
    ...
end
```

kosstennbl (commented):
Other than mentioned problem - looks nice: fixes the Tar problem and refactors the code

src/tasks/utils/cnf_manager.cr
svteb: Great catch, I reduced it to one if else block, should be alright now.

taylor (approved):

Issue #2038, opened 2024-05-17
Conflicting single_process_type and specialized_init_system checks


sysarch-repo:
**Describe the bug**
The single_process_type and specialized_init_system checks cannot be passed both in the same (e.g. cert essential) run

**To Reproduce**

$ cnf-testsuite version
CNF TestSuite version: v1.2.0

1. Execute the cert essential tests
2. See one of the tests failing. Examples for tiny and dumb-init:

🎬 Testing: [single_process_type]
resource: {kind: "Deployment", name: "dns-dserver", namespace: "default"}, pod dns-dserver-b447b56ff-4nmtx and container: dserver has more than one process type (/sbin/tini--/usr/local/bin/dns-app, /usr/local/bin/dns-app)
✖️ 🏆FAILED: [single_process_type] More than one process type used ⚖👀

🎬 Testing: [single_process_type]
resource: {kind: "StatefulSet", name: "dns-drouter", namespace: "default"}, pod dns-drouter-0 and container: drouter has more than one process type (/usr/local/bin/dumb-init--/usr/local/bin/dns-app, /usr/local/bin/dns-app)
✖️ 🏆FAILED: [single_process_type] More than one process type used ⚖👀

**Expected behavior**
The single process type check shall be made tolerant to the specialized init systems (dumb-init, tiny, s6-overlay) so that both tests can be passed in the same run.

**Device (please complete the following information):**
Linux ip-10-0-33-150 6.5.0-1020-aws https://github.com/cnti-testcatalog/testsuite/issues/20~22.04.1-Ubuntu SMP Wed May 1 16:10:50 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

**Additional context**
Note, the s6-overlay init system seems to be starting multiple processes.

martin-mat:
There seems to be unfinished implementation in k8s_kernel_introspection's verify_single_proc_tree
https://github.com/cnf-testsuite/k8s_kernel_introspection/blob/main/src/kernel_introspection/k8s.cr#L41

```
         if current_pid == original_parent_pid && ppid != "" && 
            status_name != name
            # todo exclude tini, init, dumbinit?, from violations
            Log.info { "top level parent (i.e. superviser -- first parent with different name): #{status_name}" }
            verified = false
```

PR #2054, opened 2024-05-24
Add exclusion of init system from single process check

horecoli:
## Description
This change defines a list of specialized init systems as a constant and excludes it from the verification of a single process check.
**This PR depends on https://github.com/cnf-testsuite/k8s_kernel_introspection/pull/3 which has to be merge first.**

## Issues:
Refs: #2038

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/utils/init_systems.cr
+++ b/src/tasks/utils/init_systems.cr
@@ -18,18 +18,9 @@ module InitSystems
   end
   
   def self.is_specialized_init_system?(cmd : String) : Bool
-    # Tests for tini
-    # https://github.com/krallin/tini
-    return true if cmd.includes?("tini")
-  
-    # Tests for dumb-init
-    # https://github.com/Yelp/dumb-init
-    return true if cmd.includes?("dumb-init")
-  
-    # Tests for s6-overlay
-    # https://github.com/just-containers/s6-overlay
-    return true if cmd.includes?("s6-svscan")
-  
+    SPECIALIZED_INIT_SYSTEMS.each do |init_system|
+      return true if cmd.includes?(init_system)
+    end
     # No specialized init system found
     return false
   end
--- a/shard.lock
+++ b/shard.lock
@@ -42,7 +42,7 @@ shards:
 
   k8s_kernel_introspection:
     git: https://github.com/cnf-testsuite/k8s_kernel_introspection.git
-    version: 1.0.1
+    version: 1.0.2
 
   k8s_netstat:
     git: https://github.com/cnf-testsuite/k8s_netstat.git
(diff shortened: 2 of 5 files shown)


horecoli:
The PR mentioned in the description, along with this PR, works fine with the dump-init and tini specialized init systems. However, there is different behavior in the case of the v6 init system (s6-svscan), which is not a single process init system but produces quite a lot of processes. So the main question is whether this case conflicts with the best practice of having single processes inside a container. I believe this needs to be discussed. Here is a screenshot showing how this test is failing with a list of processes:

image

martin-mat:
This needs discussion
@taylor @wavell @Smitholi67 

My opinion: It is OK to handle init process as an exception for single-process-type test/best practice. However, S6 init system itself does not follow single-process principle, so it is fine to have failing single-process check when S6 is used. And also for that, we should reconsider recommending S6 as a best-practice init system for containers.

src/tasks/utils/init_systems.cr
```diff
@@ -18,18 +18,9 @@ module InitSystems
...
-    return true if cmd.includes?("tini")
-  
-    # Tests for dumb-init
-    # https://github.com/Yelp/dumb-init
-    return true if cmd.includes?("dumb-init")
-  
-    # Tests for s6-overlay
-    # https://github.com/just-containers/s6-overlay
-    return true if cmd.includes?("s6-svscan")
-
```
kosstennbl, line 32: Comments with links to init systems could be moved to init-systems array definition to keep information there.

wavell (approved):

src/tasks/utils/init_systems.cr
horecoli: I am not sure if it is useful to keep this in the code. A better place for it would probably be somewhere in the documentation.

taylor (approved):

horecoli:
The PR - https://github.com/cnf-testsuite/k8s_kernel_introspection/pull/3 to k8s_kernel_introspection was merged and new tag was created.  So this PR is ready to merge.

martin-mat (changes requested):
shard.lock needs to be updated.

martin-mat (approved):
lgtm

tests passed:
https://github.com/cnti-testcatalog/testsuite/actions/runs/9422178235

Issue #2087, opened 2024-06-26
[Improvement] Refactor K8sTshark module


svteb:
**Is your feature request related to a problem? Please describe.**
The module currently holds 3 distinct functions, each having a duplicate with 1~ line changed. These duplicates should be revised.

**Describe the solution you'd like**
Create more sensible interfaces. Shorten the code and remove the duplicates. Currently the problem with the module is dependency on uncertain wait times (120 seconds for tshark capture). A function that would terminate tshark capture after its begun would be very helpful.

**Additional context**
The module is currently used only by suci and oran tests, thus not much code will need to be rewritten outside the module.

PR #2097, opened 2024-07-01
Feat: Rework k8s_tshark functionality and apply the changes in tests

### LINKED ISSUE #2072: [BUG] Unstable 5G tests(suci_enabled) in github actions
**Describe the bug**
5G tests fail frequently in github actions.
Example:
https://github.com/cnti-testcatalog/testsuite/actions/runs/9439225077/job/25997107636

**To Reproduce**
run github actions few times. Failures are quite frequent.

**Expected behavior**
Tests should pass if there is no error/change

svteb:
## Description
**Commit message:**

- Prior functionality was bound to fixed time of execution (120s), which introduced problems in testing (tshark session ending before the test began).
- New functionality mainly implements infinite tshark execution along with the possibility of terminating it when deemed appropriate. This is complemented with robust error handling and termination of the tshark process on unexpected crashes during initialization. NOTE: The main tests currently do not handle states where a crash could occur elsewhere and thus a hanging tshark session can still happen (although unlikely).
- The module is properly commented which should allow the user to get a quick understanding of its functionality.
- The user functionality remains the same with easier-to-comprehend function names.
- Handling of PIDs is rather problematic due to the nature of exec_by_node_bg function, which does not return the PID of the tshark process but rather the PID of the shell executing it (unverified). This is why the retrieval of PID may seem rather complicated (especially the pid_command variable). Possible solutions are listed in a comment, but these don't quite work for various reasons (globbing issues, return of incorrect PID, etc.). As for the kill -15 and kill -9 repetition, some tshark session would get stuck in a zombie state if the commands were not executed in this order.

**Additional info for reviewers:**

The spec tests that verify the correctness of this change are: `5g, oran, core`. For some reason I cannot deploy the `oran` CNF properly so the testing will probably have to be left for github actions. Some of the changes that were made removed the necessity of nesting in the main workload functions (which was thus removed). The class oriented design was somewhat necessary to allow for saving the state of the capture and its future termination. This should also resolve #2072 .

As for the code, a non-trivial amount of time was spent fine-tuning the behavior around PIDs and this is the only configuration I got to behave correctly, although I admit some parts don't look all that pretty. The main problems stem from kubectl_client which is eventually called to spawn the process, since the Process.new/run functions do not return the correct PID/print output I could not really achieve this any other way.

I also discovered that the `smf_upf_heartbeat` task essentially tests nothing.

## Issues:
Refs: #2072 #2087

## How has this been tested:
 - [x] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**
- [x] Tasks in issue are checked off

### DIFF
--- a/src/tasks/utils/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager.cr
@@ -857,8 +857,9 @@ module CNFManager
       baselines = JaegerManager.unique_services_total
       Log.info { "baselines: #{baselines}" }
     end
+
     # todo start tshark monitoring the e2 traffic 
-    tshark_log_name = ORANMonitor.start_e2_capture?(config.cnf_config)
+    capture = ORANMonitor.start_e2_capture?(config.cnf_config)
 
     # todo separate out install methods into a module/function that accepts a block
     liveness_time = 0
@@ -1053,9 +1054,10 @@ module CNFManager
     else
       tracing_used = false
     end
+
     if ORANMonitor.isCNFaRIC?(config.cnf_config)
       sleep 30 
-      e2_found = ORANMonitor.e2_session_established?(tshark_log_name)
+      e2_found = ORANMonitor.e2_session_established?(capture)
     else
       e2_found = false
     end
--- a/src/tasks/utils/k8s_tshark.cr
+++ b/src/tasks/utils/k8s_tshark.cr
@@ -5,148 +5,212 @@ require "halite"
 
 
 module K8sTshark
+  class TsharkPacketCapture
+    property capture_file_path : String
+    property pid : Int32?
+    private property node_match : JSON::Any?
 
-  def self.log_of_tshark_by_label(command, label_key, label_value, duration="120") : String
-      Log.info { "log_of_tshark_by_label command label_key label value: #{command} #{label_key} #{label_value}" }
-      all_pods = KubectlClient::Get.pods_by_nodes(KubectlClient::Get.schedulable_nodes_list)
-      pods = KubectlClient::Get.pods_by_label(all_pods, label_key, label_value) 
-      first_labeled_pod = pods[0]?
-      Log.info { "first_labeled_pod: #{first_labeled_pod}" }
-      if first_labeled_pod && first_labeled_pod.dig?("metadata", "name")
-        Log.info { "first_labeled_pod #{first_labeled_pod} metadata name: #{first_labeled_pod.dig?("metadata", "name")}" }
-        pod_name = first_labeled_pod.dig("metadata", "name")
-        Log.info { "pod_name: #{pod_name}" }
-        nodes = KubectlClient::Get.nodes_by_pod(first_labeled_pod)
-        node = nodes.first
-        #create a unique name for the log
-        # rnd = Random.new
-        # name_id = rnd.next_int
-        # tshark_log_name = "/tmp/tshark-#{name_id}.json"
-        # Log.info { "tshark_log_name #{tshark_log_name}" }
-        #
-        # #tshark -ni any  -Y nas_5gs.mm.type_id -T json 2>&1 | tee hi.log 
-        # #command= -ni any  -Y nas_5gs.mm.type_id -T json
-        # #todo check if tshark running already to keep from saturating network
-        # #todo play with reducing default duration
-        # ClusterTools.exec_by_node_bg("tshark #{command} -a duration:#{duration} 2>&1 | tee #{tshark_log_name}", node)
-        # ClusterTools.exec_by_node_bg("tshark -ni any -a duration:120 -Y nas_5gs.mm.type_id  -T json 2>&1 | tee #{tshark_log_name}", node)
-        # Log.info { "after exec by node bg" }
-        # resp = tshark_log_name
-        resp = log_of_tshark_by_node(command, node, duration)
-      else
-        resp = "label key:#{label_key} value: #{label_value} not found"
+    def initialize
+      @capture_file_path = ""
+    end
+
+    def finalize
+      if @pid
+        terminate_capture
       end
-      Log.info { "resp #{resp}" }
-      resp
-  end
-
-  def self.log_of_tshark_by_label_bg(command, label_key, label_value, duration="120") : String
-    Log.info { "log_of_tshark_by_label command label_key label value: #{command} #{label_key} #{label_value}" }
-    all_pods = KubectlClient::Get.pods_by_nodes(KubectlClient::Get.schedulable_nodes_list)
-    pods = KubectlClient::Get.pods_by_label(all_pods, label_key, label_value) 
-    first_labeled_pod = pods[0]?
-      Log.info { "first_labeled_pod: #{first_labeled_pod}" }
-    if first_labeled_pod && first_labeled_pod.dig?("metadata", "name")
-      Log.info { "first_labeled_pod #{first_labeled_pod} metadata name: #{first_labeled_pod.dig?("metadata", "name")}" }
-      pod_name = first_labeled_pod.dig("metadata", "name")
-      Log.info { "pod_name: #{pod_name}" }
-      nodes = KubectlClient::Get.nodes_by_pod(first_labeled_pod)
-      node = nodes.first
-      #create a unique name for the log
-      # rnd = Random.new
-      # name_id = rnd.next_int
-      # tshark_log_name = "/tmp/tshark-#{name_id}.json"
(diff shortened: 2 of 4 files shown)


kosstennbl (changes requested):
As was discussed online, changes for timeout design and possible nil value of "capture" variable

src/tasks/utils/oran_monitor.cr
```diff
@@ -24,49 +24,45 @@ module ORANMonitor
...
-        e2_found = true
-      else
-        Log.info { "regex not found " }
-        e2_found = false
-      end
-      Log.info { "found e2ap.successfulOutcome_element followed by e2ap.ric_ID?: #{e2_found}" }
-
-      #todo delete log file
+    
+    if capture.regex_match?(/e2ap\.successfulOutcome_element[^$]*e2ap.ric_ID/)
```
svteb, line 54: There is a missing check for capture being Nil.

src/tasks/utils/k8s_tshark.cr
```diff
@@ -5,148 +5,199 @@ require "halite"
...
+        timeout = 10 # seconds
+        waited = 0
+        while !File.exists?(pid_file) || File.size(pid_file) == 0
+          sleep 0.1
+          waited += 0.1
+  
+          if waited >= timeout
+            raise "Timeout waiting for PID file."
+          end
+        end
```
svteb, line 178: For this timeout logic, the `repeat_with_timeout` function can be used.

src/tasks/workload/5g_validator.cr
```diff
@@ -55,17 +55,13 @@ task "smf_upf_heartbeat" do |t, args|
       command = "-ni any -Y 'pfcp.msg_type == 1 or pfcp.msg_type == 2' -T json"
 
       #Baseline 
+      capture = K8sTshark::TsharkPacketCapture.new
+      capture.begin_capture_by_label(smf_key, smf_value, command)
```
svteb, line 59: I am unsure about this being a single capture and will need to review if something wrong could happen.

src/tasks/utils/cnf_manager.cr
```diff
@@ -1055,7 +1055,7 @@ module CNFManager
     end
     if ORANMonitor.isCNFaRIC?(config.cnf_config)
       sleep 30 
-      e2_found = ORANMonitor.e2_session_established?(tshark_log_name)
+      e2_found = ORANMonitor.e2_session_established?(capture.not_nil!)
```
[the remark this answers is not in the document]
svteb: Resolved

src/tasks/utils/oran_monitor.cr
svteb: Resolved

src/tasks/utils/k8s_tshark.cr
svteb: Resolved

src/tasks/workload/5g_validator.cr
svteb: It actually did not make sense for this to be a single capture, I made some changes to the task to make it somewhat sensible. **TL;DR:** As I have mentioned above in the pull request description, the `smf_upf_heartbeat` task did not measure/test anything properly. With the newest change I've pushed it might be a little more sensible.

**Long version:**

The `smf_upf_heartbeat` task is not even documented, which makes deciphering what it should do and how even harder. By reading the code and applying some speculation, the objective is probably to: 
`measure some baseline state` > `run some disruptive behavior` > `measure again during disruptive state` > `compare the two states to determine if the 5g remained "persistent"`. 

With the original implementation the "measuring" stages took around 1 second, during which nothing really was measured. I upped that to 60 for both so that we get some actual data to be compared. Upon running the test I saw that the measured counts were different before and after disruption (as expected). The spec tests pass as well. As this test is not in the essential category I think we can leave it as is for now.

src/tasks/workload/5g_validator.cr
```diff
@@ -146,33 +133,23 @@ task "suci_enabled" do |t, args|
...
+      # TODO 5g RAN (only) mobile traffic check ????
+      # use suci encyption but don't use a null encryption key
+      suci_found = capture.regex_match?(/"nas_5gs.mm.type_id": "1"/) && \
+                   !capture.regex_match?(/"nas_5gs.mm.suci.scheme_id": "0"/) && \
+                   !capture.regex_match?(/"nas_5gs.mm.suci.pki": "0"/)
+
+      Log.info { "suci_found: #{suci_found}" }
+      #todo delete log file
+
+      capture.terminate_capture
```
kosstennbl, line 152: Seeing this - i've got an idea for possible improvement:
Maybe capture could work like File.open(), take block of code and close the capture automatically after in "ensure" statement. This way it would be possible to get rid of need for terminating the captures and having "finalize" method for Capture object.

src/tasks/utils/k8s_tshark.cr
```diff
@@ -5,148 +5,198 @@ require "halite"
...
+        Log.error { error_message }
+    
+        # Attempt to kill the capture process if it was started.
+        ClusterTools.exec_by_node_bg("pkill -15 tshark && sleep 1 && pkill -9 tshark", @node_match.not_nil!)
+        @pid = nil
+    
+        raise K8sTsharkError.new(error_message)
+      ensure
+        File.delete(pid_file) if File.exists?(pid_file)
+      end
```
kosstennbl, line 193: Proposal:
```
private def retrieve_pid(command : String, pid_file : String)
  begin
    # Store the pid of the tshark process.
    ClusterTools.exec_by_node_bg(command, @node_match.not_nil!)

    # Wait for pidfile to be readable using repeat_with_timeout
    pid_found = repeat_with_timeout(timeout: 10, errormsg: "Timed out while waiting for PID file.") do
      File.exists?(pid_file) && File.size(pid_file) > 0
    end
    if pid_found
      @pid = File.read(pid_file).strip.to_i
      Log.info { "tshark process started with PID: #{@pid}" }
    else
      # Attempt to kill the capture process if it was started.
      ClusterTools.exec_by_node_bg("pkill -15 tshark && sleep 1 && pkill -9 tshark", @node_match.not_nil!)
      @pid = nil
      raise K8sTsharkError.new("Could not retrieve the PID of the tshark process")
    end
  ensure
    File.delete(pid_file) if File.exists?(pid_file)
  end
end
```
Fixed usage of repeat_with_timeout, simplified error handling and error logging.

src/tasks/utils/k8s_tshark.cr
```diff
@@ -5,148 +5,198 @@ require "halite"
 
 
 module K8sTshark
+  class TsharkPacketCapture
+    property capture_file_path : String
+    property pid : Int32?
+    private property node_match : JSON::Any?
```
kosstennbl, line 11: Idea of this object being "reusable" bothers me a little, but i don't think its too big of a problem and i can't think of a solution that would fix this and would be implementable in a reasonable time.

src/tasks/workload/5g_validator.cr
svteb: I added this option and its now used in both suci_enabled and in smf_upf_heartbeat tasks. As we discussed in a meeting, this cannot quite be applied to the oran test in `cnfmanager.cr` because the beginning and end of capture there spans some 400 lines, maybe after refactor of `sample_setup` something can be figured out.

src/tasks/utils/k8s_tshark.cr
svteb: Changed.

src/tasks/utils/k8s_tshark.cr
svteb: I also thought about this and the simplest option would be to use some flags that would effectively make the object unusable after the first capture has begun, but  I do not think it is necessary (just more bloat with if/else checks). There are warnings and error handling for the programmer to figure out if he is doing something wrong and considering that this function is not used in that many places it probably does not matter, furthermore the block `open-with` option from the above review will somehow hide this functionality. Also singleton does not work here as we can have multiple captures running at the same time (might be desired in the future).

martin-mat (approved):
lgtm

collivier (approved):

Smitholi67 (approved):
lgtm

Issue #2084, opened 2024-06-24
[Improvement] Optimize runtime of single_process_type test


martin-mat:
single_process_type takes much time. Example in GH actions:
https://github.com/cnti-testcatalog/testsuite/actions/runs/9616958098/job/26625214259

It seems that the test executes kubectl numerous times (unnecessarily?)

Issue #27, opened 2020-02-20
Sample CNFs for testing in with the CNF Conformance suite


taylor:
To help with developing and continually testing the CNF Conformance suite it would be nice to have a set of sample CNFs to use.

General requirements or guidelines for the samples:
- Easy to install (eg. helm v3.0)
- No host modifications
- Easy to install K8s extensions (if required), with no host reboot required
- Information on how the CNF is used in a real world use case (so it's purpose can be communicated)

Feel free to ideas for CNFs
-  directly to the ticket
-  more extended info to the [hackmd shared markdown](https://hackmd.io/VJV3YYtuRHqvs5_syRYj9A) (or a new one if needed)

taylor:
https://hackmd.io/VJV3YYtuRHqvs5_syRYj9A

taylor:
https://01.org/intel-data-plane-performance-demonstrators/prox-overview

petorre:
We moved PROX documentation from 01.org to OPNFV web:
- Automation scripts documentation https://wiki.opnfv.org/display/SAM/Rapid+scripting and
- code https://git.opnfv.org/samplevnf/tree/VNFs/DPPD-PROX/helper-scripts/rapid .

Latest version there is now containerized.

For NFVI Performance Characterization: @michaelspedersen and I plan to contribute it to CNF Testbed and Conformance suite.

lixuna:
New example CNF ideas may be added to the documentation at https://github.com/cncf/cnf-conformance/blob/master/EXAMPLE-CNFS.md via pull request

Closing this issue.

Issue #2103, opened 2024-07-09
[Feature] Support multiple container engines


martin-mat:
**Is your feature request related to a problem? Please describe.**
Currently, cnf-testsuite requires containerd as a k8s container engine.
https://github.com/cnti-testcatalog/testsuite/blob/v1.3.0/INSTALL.md#minimum-requirements

This limits testing and prevents platforms without containerd option to be used, for example openshift.

**Describe the solution you'd like**
Explore which container engines are commonly used (containerd, cri-o, maybe others?)
Explore container engine limitations in upstream tools the are used (litmuschaos, kyverno, kubescape etc.)
Analyze what is needed to be changed (cluster-tools, maybe additional places)

**Describe alternatives you've considered**
Keep containerd as the only supported platform engine

**How will this be tested? aka Acceptance Criteria (optional) **
Add spec tests/gh actions to run tests on platforms with all supported engines

PR #2110, opened 2024-07-17
Fix: Optimize single_process_type execution

svteb:
## Description
The single_process_type / process_check spec tests took on average 1:30:00 to execute (in github actions), this was reduced to around 10-15 minutes. It was caused by incorrect use of libraries which checked unnecessary processes multiple times.

**These two changes need to be merged first:**
cluster_tools change: [#27](https://github.com/cnf-testsuite/cluster_tools/pull/27)
k8s_kernel_introspection change: [#4](https://github.com/cnf-testsuite/k8s_kernel_introspection/pull/4)

Because this pull request makes changes in two libraries, it is rather difficult to verify it through actions. Currently there are changes to the `shards.yaml` and `shards.lock` files which will be removed once the reviewers approve the change and the individual library pull requests are merged.

**For reviewers:**
There are two big changes, the first is the replacement of `workload_resource_test` by `cnf_workload_resources`, both yield resource names, but `workload_resource_test` also yields all the container names per resource, thus we could get resources like this (it also did not return all the containers for some reason?):

```
CNFManager.workload_resource_test(args, config) do |resource, container, initialized|

resource: {kind: "Deployment", name: "nginx-webapp", namespace: "cnfspace"}
resource: {kind: "Deployment", name: "nginx-webapp", namespace: "cnfspace"}
resource: {kind: "Deployment", name: "nginx-webapp", namespace: "cnfspace"}
resource: {kind: "Pod", name: "sidecar-container-demo", namespace: "cnfspace"}
resource: {kind: "Pod", name: "sidecar-container-demo", namespace: "cnfspace"}
```

optimized version:

```
CNFManager.cnf_workload_resources(args, config) do |resource|

resource: {kind: "Deployment", name: "nginx-webapp", namespace: "cnfspace"}
resource: {kind: "Pod", name: "sidecar-container-demo", namespace: "cnfspace"}
```

With this change, we do not iterate over every resource multiple times. The second big change is the addition of pid filtering by cgroups. The previous version was incorrectly checking processes on node instead of a container. This lead to something like this (which makes no sense considering we want verify to processes of specific container).

```
for each container:
         container_node = get_container_node()
         get_all_pids_on_node(container_node)
```

optimized version:

```
for each container:
         container_node = get_container_node()
         get_all_pids_on_container(container, container_node)
```

The cgroup filtering currently depends on `ctr` purely because it was easier to write than `runc` filtering. Considering the fact that the current testsuite depends on containerd runtime (`ctr`) it should not be too problematic and can be changed in the future. Finally, I noticed that the cluster_tools library provided a function that was a copy of the code in `single_process_check`. That same function is also shared by the zombie task which has some unique quirks I will not get too deep into here. That shared function (`ClusterTools.all_containers_by_resource?`) had to be changed slightly so it would not break the zombie task. A better refactor can be considered in the future but it is outside the scope of this pull request.

## Issues:
Refs: #2084

## How has this been tested:
 - [x] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -9,8 +9,7 @@ shards:
     version: 1.3.1
 
   cluster_tools:
-    git: https://github.com/cnf-testsuite/cluster_tools.git
-    version: 1.0.5
+    git: https://github.com/svteb/cluster_tools.git
 
   commander:
     git: https://github.com/mrrooijen/commander.git
@@ -41,12 +40,10 @@ shards:
     version: 0.9.0+git.commit.f62bfcdfbe65ee31c46c3d9951cee08ac2e0bee0
 
   k8s_kernel_introspection:
-    git: https://github.com/cnf-testsuite/k8s_kernel_introspection.git
-    version: 1.0.2
+    git: https://github.com/svteb/k8s_kernel_introspection.git
 
   k8s_netstat:
-    git: https://github.com/cnf-testsuite/k8s_netstat.git
-    version: 1.0.1
+    git: https://github.com/svteb/k8s_netstat.git
 
   kernel_introspection:
     git: https://github.com/cnf-testsuite/kernel_introspection.git
--- a/shard.lock
+++ b/shard.lock
@@ -9,7 +9,8 @@ shards:
     version: 1.3.1
 
   cluster_tools:
-    git: https://github.com/svteb/cluster_tools.git
+    git: https://github.com/cnf-testsuite/cluster_tools.git
+    version: 1.0.6
 
   commander:
     git: https://github.com/mrrooijen/commander.git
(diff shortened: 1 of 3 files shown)


kosstennbl (commented):
Nice change, optimizations for this test are very welcome.
Few additional points:
- Maybe we could use crictl for k8s introspection change for easier support of #2103 in case it will be implemented?
- Operation of workload_resource_test and cnf_workload_resources is complicated and feels that workload_resource_test design could be a cause for multiple slowdowns in our tests. Probably we should look into fixing this at a higher scale. Not a scope of this PR though.

svteb:
> Nice change, optimizations for this test are very welcome. Few additional points:
> 
> * Maybe we could use "crictl" for k8s introspection change for easier support of [[Feature] Support multiple container engines #2103](https://github.com/cnti-testcatalog/testsuite/issues/2103) in case it will be implemented?
> * Operation of workload_resource_test and cnf_workload_resources is complicated and feels that workload_resource_test design could be a cause for multiple slowdowns in our tests. Probably we should look into fixing this at a higher scale. Not a scope of this PR though.

Yes, after some discussions we have come to the conclusion that `ctr` might not be generic enought. The issue is that every container engine has some nuances that prevent easy access to cgroups/filtering pids. We've decided to use `runc` instead which is a lower level engine. The issue with `crictl` is that it is too high-level and does not provide access to cgroups.

kosstennbl (approved):
lgtm

taylor (approved):

PR #2113, opened 2024-07-25
workflow: Cancel workflows on push

kosstennbl:
## Description

In our CICD, when new commit is pushed (or amended) while workflow for old one is still running - there is no auto-cancelation. Old workflows have to be canceled manually by maintainers to avoid unneccessary waiting.

This change fixes that issue by grouping workflows by PR number or branch ref (in case for push event), and if new workflow is ran in the same group - old one would be automatically cancelled.

Change is relatively simple, so issue wasn't created for it.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -12,6 +12,11 @@ on:
       - '!**.md'
       - '!docs/*'
       - '!doc-lint/*'
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Smitholi67 (approved):

svteb (approved):
LGTM

horecoli (approved):
Very useful thing :)

Issue #2083, opened 2024-06-24
[BUG] spec tests don't fail on unsuccessful cnf-setup


martin-mat:
**Describe the bug**
In case that the cnf setup is unsuccessful on issues like:
- cluster-tools not ready after waiting for it (180)
- deployment of the cnf returns an error
- deployment of cnf not ready after waiting for it (180)
spec test ignore it and continue in test execution

see #2081 

**Expected behavior**
Spec test will fail as soon as mentioned errors occur

martin-mat:
#1937

kosstennbl:
#2114 Merged

Issue #1937, opened 2024-03-26
[BUG] cnf_setup command is not working correctly in case of failing cnf creation


horecoli:
**Describe the bug**
It looks like that cnf_setup doesn't work correctly. It is finishing successfully even if cnf is not created or it is not in running state.
Example with envoy:
`./cnf-testsuite cnf_setup cnf-config=example-cnfs/envoy/cnf-testsuite.yml`
`Successfully created directories for cnf-testsuite`
`cnf-testsuite namespace already exists on the Kubernetes cluster`
`cnf setup online mode`
`Successfully setup envoy`
`cnf setup online mode complete`

`ubuntu@vm:~/test/testsuite$ kubectl get pods`
`No resources found in default namespace.`

Example coredns:
`./cnf-testsuite cnf_setup cnf-config=example-cnfs/coredns/cnf-testsuite.yml`
`Successfully created directories for cnf-testsuite`
`cnf-testsuite namespace already exists on the Kubernetes cluster`
`cnf setup online mode`
`Successfully setup coredns`
`cnf setup online mode complete`

`ubuntu@vm:~/git/cnf-testsuite$ kubectl get pods`
`NAME                               READY   STATUS             RESTARTS   AGE`
`coredns-coredns-7b6b6c5d6d-4hlvk   0/1     ImagePullBackOff   0          3m59s`

**To Reproduce**
I found out this issue when I had some issues with image pulling. So for reproduction is needed little bit broken environment to have issues with pulling (maybe wrong dns setting).

**Expected behavior**
cnf_setup command should fail or at least return correct exit status.

**Device (please complete the following information):**

- OS [e.g. Linux, iOS, Windows, Android]
- Distro [e.g. Ubuntu]
- Version [e.g. 18.04]
- Architecture [e.g. x86, arm]

**How will this be tested? aka Acceptance Criteria (optional)**

(optional: unnecessary for things like spelling errors and such)

Once this issue is address how will the fix be verified?

**Additional context**

kosstennbl:
#2114 Merged

PR #2114, opened 2024-07-29
cnf_setup: Exit on setup failure or timeout

kosstennbl:
## Description
Fix handling unsuccessful cnf setup, if resources are not ready after timeout - then setup should fail with exit 1. 
Change specs accordingly, replace wait_count=0 with new option "ignore_setup_timeout".
Add cnf_setup function to ShellCmd and use it for installation of CNFs during specs.

## Issues:
Refs: #1937 #2083


src/tasks/utils/cnf_manager.cr
```diff
@@ -815,12 +820,25 @@ module CNFManager
...
+      if !skip_wait_for_install
+        stdout_success "Waiting for resource availability, timeout for each resource is #{wait_count} seconds \n"
+        workload_resource_names = resource_names.select { |resource| 
+        ["replicaset", "deployment", "statefulset", "pod", "daemonset"].includes?(resource[:kind].downcase) 
+      }
+        total_resource_count = workload_resource_names.size()
+        current_resource_number = 1
+        workload_resource_names.each do | resource |
+          # \e[1A\e[K is a sequence that allows to clear current line and place cursor at its beginning
+          stdout_success "\e[1A\e[K" + "Waiting for resource (#{current_resource_number}/#{total_resource_count}): #{resource[:kind]} #{resource[:name]}"
```
martin-mat, line 832: I'd suggest to do 
```
stdout_success  "\e[K" + "Waiting for resource (#{current_resource_number}/#{total_resource_count}): #[{resource[:kind]}] #{resource[:name]}" + "\e[1A"
```
and not to do extra new line on 824
and print extra "\e[K" after the loop. 

Advantages of such modifications:
- this will not leave last status line on console. Leaving is somehow not consistent with idea having this just as an intermediate console info (and anyway it is reported again on 835 in case of error)
- on screens without ansi support (and when redirected to a file) no unnecessary extra line will be present after printout at 824

Maybe, even better, and to be usable also on other places, we can define common functions
stdout_update_line(str) -> stdout_success("\e[K" + str + "\e[1A")
stdout_cleanup_line() -> stdout_success("\e[K")

Also, marking the kind of the resource would help, maybe square brackets?
```
Waiting for resource (1/1): [Deployment] coredns-coredns
```

martin-mat (changes requested):

src/tasks/utils/cnf_manager.cr
kosstennbl: Suggestion haven't worked as expected, "\e[1A" in the end of the string moves the cursor, but next print that doesn't clear the line works incorrectly (it writes on top of current text). Possible solution is doing cleanup_line before each next print, but i don't like that solution. I think that each print should leave the terminal in a stable state. 

Check current solution, it writes "All CNF resources are up!" by replacing last "Waiting for resource" line and has separate function for prints like these.

martin-mat: looks ok, thanks.

src/tasks/utils/utils.cr
```diff
@@ -411,6 +411,10 @@ def stdout_success(msg)
   puts msg.colorize(:green)
 end
 
+def stdout_success_same_line(msg)
```
martin-mat, line 414: maybe better would be to add a new param to stdout_success(msg, same_line=false)

src/tasks/utils/utils.cr
```diff
@@ -419,6 +423,10 @@ def stdout_failure(msg)
   puts msg.colorize(:red)
 end
 
+def stdout_failure_same_line(msg)
```
martin-mat, line 426: dtto

src/tasks/utils/cnf_manager.cr
```diff
@@ -815,12 +820,25 @@ module CNFManager
...
-          KubectlClient::Get.resource_wait_for_install(resource[:kind], resource[:name], wait_count: wait_count, namespace: resource[:namespace])
+      if !skip_wait_for_install
+        stdout_success "Waiting for resource availability, timeout for each resource is #{wait_count} seconds \n"
+        workload_resource_names = resource_names.select { |resource| 
+        ["replicaset", "deployment", "statefulset", "pod", "daemonset"].includes?(resource[:kind].downcase) 
+      }
+        total_resource_count = workload_resource_names.size()
+        current_resource_number = 1
+        workload_resource_names.each do | resource |
+          # \e[1A\e[K is a sequence that allows to clear current line and place cursor at its beginning
```
martin-mat, line 831: remove this comment

src/tasks/utils/cnf_manager.cr
```diff
@@ -815,12 +820,25 @@ module CNFManager
           end
         end
       end
-
-      resource_names.each do | resource |
-        case resource[:kind].downcase
-        when "replicaset", "deployment", "statefulset", "pod", "daemonset"
-          KubectlClient::Get.resource_wait_for_install(resource[:kind], resource[:name], wait_count: wait_count, namespace: resource[:namespace])
+      if !skip_wait_for_install
+        stdout_success "Waiting for resource availability, timeout for each resource is #{wait_count} seconds \n"
```
martin-mat, line 824: just nitpicking - better if this \n is done as an escape sequence to be consistent with what it prepares for. And it will not print excess newline in case that the terminal does not support them.

src/tasks/utils/cnf_manager.cr
```diff
@@ -815,12 +820,25 @@ module CNFManager
...
+      }
+        total_resource_count = workload_resource_names.size()
+        current_resource_number = 1
+        workload_resource_names.each do | resource |
+          # \e[1A\e[K is a sequence that allows to clear current line and place cursor at its beginning
+          stdout_success_same_line "Waiting for resource (#{current_resource_number}/#{total_resource_count}): [#{resource[:kind]}] #{resource[:name]}"
+          ready = KubectlClient::Get.resource_wait_for_install(resource[:kind], resource[:name], wait_count: wait_count, namespace: resource[:namespace])
+          if !ready
+            stdout_failure_same_line "CNF setup has timed-out, [#{resource[:kind]}] #{resource[:name]} is not ready after #{wait_count} seconds."
+            stdout_failure "Preffered course of actions would be to investigate the resource in cluster, then call cnf_cleanup and try to reinstall the CNF."
```
martin-mat, line 836: Preffered -> Recommended

martin-mat (changes requested):

martin-mat:
It looks that this change did its job and revealed issues with cnf setup in spec tests. They need to be addressed though.

src/tasks/utils/cnf_manager.cr
kosstennbl: This will require creating additional function or parameter for printing with newline before carriage return and line clear. It seems excess to me.

src/tasks/utils/cnf_manager.cr
kosstennbl: Moved appropriately

src/tasks/utils/utils.cr
kosstennbl: Done

src/tasks/utils/utils.cr
kosstennbl: Done

src/tasks/utils/cnf_manager.cr
kosstennbl: Done

kosstennbl:
Issues with "envoy" chart are brought by #2116, they should be resolved for CI to pass.

martin-mat:
the change looks fine for me now, I will approve after the envoy issue is addressed.

martin-mat:
rebase pls

src/tasks/utils/cnf_manager.cr
```diff
@@ -802,12 +806,24 @@ module CNFManager
...
+        ["replicaset", "deployment", "statefulset", "pod", "daemonset"].includes?(resource[:kind].downcase) 
+      }
+        total_resource_count = workload_resource_names.size()
+        current_resource_number = 1
+        workload_resource_names.each do | resource |
+          stdout_success "Waiting for resource (#{current_resource_number}/#{total_resource_count}): [#{resource[:kind]}] #{resource[:name]}", same_line: true
+          ready = KubectlClient::Get.resource_wait_for_install(resource[:kind], resource[:name], wait_count: wait_count, namespace: resource[:namespace])
+          if !ready
+            stdout_failure "CNF setup has timed-out, [#{resource[:kind]}] #{resource[:name]} is not ready after #{wait_count} seconds.", same_line: true
+            stdout_failure "Recommented course of actions would be to investigate the resource in cluster, then call cnf_cleanup and try to reinstall the CNF."
```
martin-mat, line 821: Recommended

martin-mat (changes requested):

martin-mat (approved):
lgtm

svteb (approved):
lgtm

Issue #2115, opened 2024-07-30
[MAINTENANCE] Remove and exchange duplicate test for privileged containers


horecoli:
In the current version of the testsuite, there are 2 tests for privileged containers. They are:

- The "in-house" implementation (privileged) which allows for the exclusion of certain containers from the test.
- The Kubescape solution (privileged_containers) which lacks this functionality to exclude containers from the test.

Since privileged_containers is an essential test, it would be better to replace it with the privileged test, which fulfills the requirements including container exclusion. The best solution would be to swap the names of these two tests and remove the Kubescape implementation, at least until container exclusion functionality is implemented in the Kubescape repository.

PR #2119, opened 2024-07-31
Removal and exchange duplicate test for privileged containers

horecoli:
## Description
This change removes the Kubescape implementation of test for privileged containers. The reason behind this decision is the absence of functionality to exclude containers from the test. Instead, it has been replaced with an in-house implementation of this test, which has been renamed from "privileged" to "privileged_containers".

## Issues:
Refs: #2115

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


spec/workload/security_spec.cr
```diff
@@ -4,35 +4,35 @@ require "../../src/tasks/utils/utils.cr"
...
 
-  it "'privileged' should pass with a non-privileged cnf", tags: ["privileged"]  do
+  it "'privileged_containers' should pass with a non-privileged cnf", tags: ["privileged"]  do
     begin
       result = ShellCmd.run_testsuite("cnf_setup cnf-config=sample-cnfs/sample-statefulset-cnf/cnf-testsuite.yml")
       Log.debug { result[:output] }
-      result = ShellCmd.run_testsuite("privileged verbose")
+      result = ShellCmd.run_testsuite("privileged_containers verbose")
       result[:status].success?.should be_true
       (/Found.*privileged containers.*coredns/ =~ result[:output]).should be_nil
```
kosstennbl, line 13: This output check doesn't seem right when comparing to test outputs, the test will always pass.

spec/workload/security_spec.cr
```diff
@@ -4,35 +4,35 @@ require "../../src/tasks/utils/utils.cr"
 
 describe "Security" do
 
-  it "'privileged' should pass with a non-privileged cnf", tags: ["privileged"]  do
+  it "'privileged_containers' should pass with a non-privileged cnf", tags: ["privileged"]  do
```
kosstennbl, line 7: Maybe tag could be changed to privileged_containers? And same would be used for privilege escalation tests.

kosstennbl (changes requested):
TEST_DOCUMENTATION.yml requires a small change.
utils_spec.cr:181-183, seems that prefix is also needed to be changed.

spec/workload/security_spec.cr
horecoli: no, I don't think so, it was same also before without changing tag. Tag is probably used for all privileged tests including escalation.

spec/workload/security_spec.cr
horecoli: I am not sure what you mean, I didn't change anything related to output, I just renamed test during its execution.

.github/workflows/actions.yml
```diff
@@ -402,7 +402,7 @@ jobs:
         ./cnf-testsuite setup
         wget -O cnf-testsuite.yml https://raw.githubusercontent.com/cnti-testcatalog/testsuite/main/example-cnfs/coredns/cnf-testsuite.yml
         ./cnf-testsuite cnf_setup cnf-config=./cnf-testsuite.yml
-        LOG_LEVEL=info ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~privileged ~increase_capacity ~decrease_capacity ~install_script_helm ~helm_chart_valid ~helm_chart_published verbose
+        LOG_LEVEL=info ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~privileges ~increase_capacity ~decrease_capacity ~install_script_helm ~helm_chart_valid ~helm_chart_published verbose
```
kosstennbl, line 405: I think task names are used there, not spec tags. Should be ~privileged_containers

.github/workflows/actions.yml
```diff
@@ -477,7 +477,7 @@ jobs:
         ./cnf-testsuite setup
         wget -O cnf-testsuite.yml https://raw.githubusercontent.com/cnti-testcatalog/testsuite/main/example-cnfs/coredns/cnf-testsuite.yml
         ./cnf-testsuite cnf_setup cnf-config=./cnf-testsuite.yml
-        LOG_LEVEL=info ./cnf-testsuite all ~resilience ~compatibility ~pod_network_latency ~platform ~privileged ~increase_capacity ~decrease_capacity ~ip_addresses ~liveness ~readiness ~rolling_update ~rolling_downgrade ~rolling_version_change ~nodeport_not_used ~hostport_not_used ~hardcoded_ip_addresses_in_k8s_runtime_configuration ~install_script_helm ~helm_chart_valid ~helm_chart_published ~rollback ~secrets_used ~immutable_configmap verbose
+        LOG_LEVEL=info ./cnf-testsuite all ~resilience ~compatibility ~pod_network_latency ~platform ~privileges ~increase_capacity ~decrease_capacity ~ip_addresses ~liveness ~readiness ~rolling_update ~rolling_downgrade ~rolling_version_change ~nodeport_not_used ~hostport_not_used ~hardcoded_ip_addresses_in_k8s_runtime_configuration ~install_script_helm ~helm_chart_valid ~helm_chart_published ~rollback ~secrets_used ~immutable_configmap verbose
```
kosstennbl, line 480: ditto

.github/workflows/actions.yml
horecoli: No, there was just "privileged" before, which was tag  and not a task name, that was the reason why I changed that only in last commit when I changed tag.

.github/workflows/actions.yml
horecoli: done

spec/workload/security_spec.cr
horecoli: but yes, there was issue with regex before, which we didn't know about it. I fixed it in last commit. Thanks.

kosstennbl (approved):
LGTM

martin-mat (approved):
lgtm

rich-l (approved):
lgtm

Issue #2117, opened 2024-07-31
[MAINTENANCE] Align cert set with "Certification 2.0 beta" - reduce it to 19 essential tests


martin-mat:
Beta 2.0 certification criterion consists of 19 essential tests. Align accordingly execution of "cnf-testsuite cert" command (it should execute only those 19 essential tests)

lfn-cnti/certification#154
lfn-cnti/certification#155

PR #2126, opened 2024-08-05
cert: remove cert tag from non-essential tests

kosstennbl:
## Description
Certification was updated to count only for essential tests. 
This commit reflects the change in certification.

## Issues:
Refs: #2117

### DIFF
--- a/embedded_files/points.yml
+++ b/embedded_files/points.yml
@@ -9,32 +9,32 @@
 
 - name: reasonable_image_size
   emoji: "⚖👀"
-  tags: [microservice, dynamic, workload, cert, normal]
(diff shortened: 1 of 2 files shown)


martin-mat:
just to document: "bonus tests should not be includded in the maximum points when a failure occurs" removed because:
- with current points.yml does not make sense
- the test does not make a good sense anyway (checking total number of tests)
Thus removing the spec test.

martin-mat (approved):
lgtm

rich-l (approved):
lgtm

svteb (approved):
lgtm

Issue #2130, opened 2024-08-09
[Feature] CNF Installation (2), Config format transformer for older version


svteb:
2nd step of #2120.

The changes to config format proposed in #2121 would make the pre-existing configs of users obsolete, forcing the users to rewrite them.  To avoid this hassle a transformer will be implemented that will transform the old format to the newer one.

**Possible solutions:**
1. Read the old config file -> Store its contents in a data structure -> Create a new config file from it.
Pros:
-- Users will get a tangible new-version config file.
-- Easier to debug and look for errors.
-- This could be a quick way to update all the sample/example-cnfs.
Cons:
-- Creating the new file is possibly unnecessary and will add extra steps.
-- The location of the new config file is questionable upon creation.
2. Read the old config file -> Store its contents in the new data structure.
Pros:
-- Possibly less code than 1. option.
Cons:
-- Less tangible results.
-- Harder debugging.

We had a short discussion and leaned more towards the 2. option. It is also possible that there is a simple way to create a new config file once the contents are stored in the new data structure if some users desire it (I have not explored this yet).

As of now the structure of the new config file has not been confirmed (draft in #2129).

Issue #2120, opened 2024-07-31
[EPIC] Redesign CNF installation process


kosstennbl:
[EPIC] Redesign CNF installation process

Current installation process is quite limiting and hard to maintain. Multiple issues are already created and they only build up: #2095, #2106, #2036.

The idea is to redesign CNF installation process in a way that fixes some of the current issues and makes it easier to expand installation functionality (and possibly, in the future, make addition of new installation methods much easier).

Central idea of this change is to install CNFs in different ways, but then gather all installed resources as manifest files. Then, during testing, all tests would have to work only with manifest files regardless of installation method of CNF.

This planned redesign seems to be quite complex and time-consuming, so better approach would be to split it into several different issues and PRs, the following is the plan for the change (up to discussion, can be changed):

1. Implement new cnf-testsuite.yml config format and its parsing (#2121).
2. Implement additional functionality tied to configuration.
    1. Implement a way to transform old configs into new format.
    2. Implement generate_config for new format.
3. Rewrite all sample/example CNF configs and change default config format to new one.
4. Implement a way to create common manifest, use it for all tests that require manifest (#2125).
5. Redesign tests that are not compliant with the common manifest or require deployment type-specific functionality to function.
6. Prepare new installation process (and multiple charts support as a part of this).
7. Check how specs would be affected, prepare for required fixes.
8. Switch to new installation (use all prepared code), fix tests, remove old code.

It is planned to create GitHub issues and link them there when they would be ready for start.
This issue should be resolved as soon as all connected issues and tasks are completed.

svteb:
Finally finished! The last pull request #2171 has been successfully merged.

PR #2133, opened 2024-08-12
config: CNF installation (2.1) Transformer for old version of config

svteb:
## Description
The transformer transforms the old config according to the structure proposed in #2129.

**Notes**
1. To allow for some input validation, the v1 config was reimplemented as class in `config_v1.cr`.
2. Certain arguments are found in cnf-testsuite.yml files spread across the samples and examples that are unused anywhere in the code, these are still parsed but ignored to avoid too much manual intervention in later stages.
3. A new task has been added to allow for transformation, users can execute it by calling this command:

`./cnf-testsuite transform_config OLD_FILE_PATH NEW_FILE_PATH`

4. The `ConfigTransformer` class found in `config_transformer.cr` automatically detects the version of the old config and transforms it to the latest version (currently only v1 -> v2), but allows for future extendibility.
5. Upon determining the version of the old config, the main `transform` function is called which does the actual transformation in code (through use of Hashes and YAML::Any).
6. The `transform` function uses the appropriate transformation rules (`V1ToV2Transformation class` in `v1_to_v2_transformation.cr` file)
7. To make future extension of transformation rules easier, the two functions `hash_to_yaml_any` and `remove_nil_values` have been added (`Transformation class` in `transformation.cr` file), these convert the underlying hashes and arrays to the `YAML::Any` type and remove any nil branches so they don't clutter the final output.
8. After the transformation is done the result can be printed to either String or dumped to a file through the `serialize_to_string` and `serialize_to_file` functions.
9. Most edge cases have been resolved through the use of exceptions and their catching, giving users a readable output in case of any errors.

## Issues:
Refs:  #2130

## How has this been tested:
 - [x] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] My change requires a change to the documentation.
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/utils/cnf_installation/transformer/config_transformer.cr
+++ b/src/tasks/utils/cnf_installation/transformer/config_transformer.cr
@@ -1,0 +1,104 @@
+require "yaml"
+
+module CNFInstall
+  module Config
+    # Base class for configuration
+    # WILL BE REBASED ELSEWHERE
+    class ConfigBase
+      include YAML::Serializable
+      include YAML::Serializable::Strict
+    end
+
+    # REQUIRES FUTURE EXTENSION in case of new config format.
+    enum ConfigVersion
+      V1
+      V2
+      Latest = V2
+    end
+
+    class ConfigTransformer
+      @new_config : YAML::Any
+      @old_config : ConfigBase
+      @version : ConfigVersion
+
+      # The initializer reads the config file twice to determine the config version 
+      # of @old_config, this cannot be removed due to later parsing to appropriate class.
+      def initialize(old_config_path : String)
+        tmp_content = File.read(old_config_path)
+        yaml_content = YAML.parse(tmp_content).as_h
+
+        # Automatic version detection to streamline the transformation
+        @version = detect_version(yaml_content)
+        if @version == ConfigVersion::Latest
+          stdout_warning "Your config is the latest version."
+          exit(1)
+        end
+
+        @new_config = YAML::Any.new({} of YAML::Any => YAML::Any)
+        @old_config = parse_old_config(tmp_content)
+      end
+
+      # Serialize the transformed config to a string.
+      def serialize_to_string : String
+        YAML.dump(@new_config)
+      end
+
+      # Serialize the transformed config to a file and return the file path.
+      def serialize_to_file(file_path : String) : String
+        File.write(file_path, serialize_to_string)
+        file_path
+      end
+
+      # Detects the config version.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      private def detect_version(yaml_content : Hash(YAML::Any, YAML::Any)) : ConfigVersion
+        version_value = yaml_content["config_version"]?.try(&.to_s)
+
+        if version_value
+          begin
+            ConfigVersion.parse(version_value.upcase)
+          rescue ex : Exception
+            raise UnsupportedConfigVersionError.new(version_value)
+          end
+        else
+          ConfigVersion::V1
+        end
+      end
+
+      # Parses the config to the correct class.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      private def parse_old_config(tmp_content : String) : ConfigBase
+        begin
+          case @version
+          when ConfigVersion::V1
+            ConfigV1.from_yaml(tmp_content)
+          else
+            raise UnsupportedConfigVersionError.new(@version)
+          end
+        rescue ex : YAML::ParseException
+          # This is usually raised in case of unexpected YAML keys.
+          stdout_failure "Failed to parse config: #{ex.message}."
+          stdout_failure "Please check your YAML fields for correctness."
+          exit(1)
+        end
+      end
+
+      # Performs the transformation from V1 to V2.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      def transform
+        case @version
+        when ConfigVersion::V1
+          @new_config = V1ToV2Transformation.new(@old_config.as(ConfigV1)).transform
+        else
+          raise UnsupportedConfigVersionError.new(@version)
+        end
+      end
+    end
+
+    class UnsupportedConfigVersionError < Exception
+      def initialize(version : ConfigVersion | String)
+        super "Unsupported configuration version: #{version.is_a?(ConfigVersion) ? version.to_s.downcase : version}"
+      end
+    end
+  end
+end    
--- a/src/tasks/utils/cnf_installation/transformer/transformation_base.cr
+++ b/src/tasks/utils/cnf_installation/transformer/transformation_base.cr
@@ -1,0 +1,37 @@
+module CNFInstall
+  module Config
+    # The rules need to be somewhat explicit, different approaches have been attempted
+    # but due to crystals strict typing system they have not been viable/would be too complicated.
+    # 
+    # In case of future extension, create a new transformation rules class (VxToVyTransformation),
+    # This class should inherit the TransformationBase class and make use of process_data
+    # function at the end of its transform function.
+    class TransformationBase
+      @new_config : YAML::Any
+  
+      def initialize(@old_config : ConfigV1)
+        @new_config = YAML::Any.new({} of YAML::Any => YAML::Any)
+      end
+  
+      # Recursively remove any empty hashes/arrays/values and convert data to YAML::Any.
+      private def process_data(data : Hash | Array | String | Nil) : YAML::Any?
+        case data
+        when Array
+          processed_array = data.map { |item| process_data(item) }.compact
+          processed_array.empty? ? nil : YAML::Any.new(processed_array)
+        when Hash
+          processed_hash = Hash(YAML::Any, YAML::Any).new
+          data.each do |k, v|
+            processed_value = process_data(v)
+            processed_hash[YAML::Any.new(k)] = processed_value unless processed_value.nil?
+          end
+          processed_hash.empty? ? nil : YAML::Any.new(processed_hash)
+        when String
+          YAML::Any.new(data)
+        else
+          nil
+        end
+      end
+    end
+  end
+end
--- a/src/tasks/utils/cnf_installation/transformer/v1_to_v2_transformation.cr
+++ b/src/tasks/utils/cnf_installation/transformer/v1_to_v2_transformation.cr
@@ -1,0 +1,120 @@
+module CNFInstall
+  module Config
+    # Rules for configV1 to configV2 transformation
+    class V1ToV2Transformation < TransformationBase
+      def transform : YAML::Any
+        new_config_hash = {
+          "config_version" => "v2",
+          "common_parameters" => transform_common_parameters,
+          "dynamic_parameters" => transform_dynamic_parameters,
+          "deployments" => transform_deployments,
+        }
+      
+        # Convert the entire native hash to stripped YAML::Any at the end.
+        @new_config = process_data(new_config_hash).not_nil!
+      end
+      
+      private def transform_common_parameters : Hash(String, Array(Hash(String, String | Nil)) | Array(String) | Hash(String, String | Nil))
+        common_params = {} of String => Array(Hash(String, String | Nil)) | Array(String) | Hash(String, String | Nil)
+      
+        common_params = {
+          "white_list_container_names" => @old_config.white_list_container_names,
+          "docker_insecure_registries" => @old_config.docker_insecure_registries,
+          "image_registry_fqdns" => @old_config.image_registry_fqdns,
+          "container_names" => transform_container_names,
+          "five_g_parameters" => transform_five_g_parameters
+        }.compact
+      
+        common_params
+      end
+      
+      private def transform_container_names : Array(Hash(String, String | Nil))
+        if @old_config.container_names
+          containers = @old_config.container_names.not_nil!.map do |container|
+            {
+              "name" => container.name,
+              "rollback_from_tag" => container.rollback_from_tag,
+              "rolling_update_test_tag" => container.rolling_update_test_tag,
+              "rolling_downgrade_test_tag" => container.rolling_downgrade_test_tag,
+              "rolling_version_change_test_tag" => container.rolling_version_change_test_tag
+            }
+          end
+  
+          return containers
+        end
+  
+        [] of Hash(String, String | Nil)
+      end
+  
+      private def transform_dynamic_parameters : Hash(String, String | Nil)
+        {
+          "source_cnf_dir" => @old_config.source_cnf_dir,
+          "destination_cnf_dir" => @old_config.destination_cnf_dir
+        }
+      end
+      
+      private def transform_deployments : Hash(String, Array(Hash(String, String | Nil)))
+        deployments = {} of String => Array(Hash(String, String | Nil))
+  
+        if @old_config.manifest_directory
+          deployments["manifests"] = [{
+            "name" => @old_config.release_name,
+            "manifest_directory" => @old_config.manifest_directory
+          }]
+        elsif @old_config.helm_directory
+          deployments["helm_directories"] = [{
+            "name" => @old_config.release_name,
+            "helm_directory" => @old_config.helm_directory,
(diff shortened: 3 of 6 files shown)


martin-mat:
Can the architecture be made more future-proof towards possible future config versions?
- Identify source version of the confing on the input
- Have possibility to tell which config version one needs to transform to (perhaps with hardcoded config version actually used by the testsuite as a constant somewhere)
- Make an easy way for future introduction of config v3.0 and newer, with possibility to transform from 1.0 and 2.0

martin-mat:
Additionally, how about error handling? Like non-supported config keys on input? Maybe other error conditions?

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+      container_names.map do |container|
+        {
+          "name" => container["name"]?.as_s,
+          "rolling_update_test_tag" => container["rolling_update_test_tag"]?.as_s,
+          "rolling_downgrade_test_tag" => container["rolling_downgrade_test_tag"]?.as_s,
+          "rolling_version_change_test_tag" => container["rolling_version_change_test_tag"]?.as_s,
+          "rollback_from_tag" => container["rollback_from_tag"]?.as_s
+        }
+      end
+    end
```
kosstennbl, line 65: Maybe it is possible to take the whole list from container_names key and copy it somehow?

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+        "routingIndicator" => @old_config["routingIndicator"]?.as_s,
+        "enabled" => @old_config["enabled"]?.as_s,
+        "count" => @old_config["count"]?.as_s,
+        "initialMSISDN" => @old_config["initialMSISDN"]?.as_s,
+        "key" => @old_config["key"]?.as_s,
+        "op" => @old_config["op"]?.as_s,
+        "opType" => @old_config["opType"]?.as_s,
+        "type" => @old_config["type"]?.as_s,
+        "apn" => @old_config["apn"]?.as_s,
+        "emergency" => @old_config["emergency"]?.as_s
```
kosstennbl, line 91: These parameters are under fiveG_core hash. I think they should be accessible as @old_config["fiveG_core"]["amf_service_name"]

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+
+    private def transform_install_method(install_method : Tuple?) : Hash(String, String | Nil)
+      return nil unless install_method
+    
+      # Not sure about this one
+      {
+        "type" => install_method[0]?.as_s,
+        "description" => install_method[1]?.as_s
+      }
+    end
```
kosstennbl, line 103: Install method is not needed to be transferred, v2 config uses this parameter differently (or won't use at all).
(I should have mentioned that in the issue, but it wasn't clear at the moment)

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+
+    private def transform_helm_charts : Array
+      [{
+        "name" => @old_config["release_name"]?.as_s,
+        "helm_repo_name" => @old_config["release_name"]?.as_s,
+        "helm_repo_url" => @old_config["release_name"]?.as_s,
+        "helm_chart_name" => @old_config["release_name"]?.as_s,
+        "helm_values" => @old_config["release_name"]?.as_s,
+        "namespace" => @old_config["release_name"]?.as_s
```
kosstennbl, line 112: Should be unpacked differently:
"helm_repo_name" => @old_config["helm_repository"]["name"]?.as_s
"helm_repo_url" => @old_config["helm_repository"]["repo_url"]?.as_s,
...

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+      }]
+    end
+
+    private def transform_helm_directories : Array
+      [{
+        "name" => @old_config["release_name"]?.as_s,
+        "helm_directory" => @old_config["helm_directory"]?.as_s,
+        "namespace" => @old_config["namespace"]?.as_s
```
kosstennbl, line 120: old - helm_install_namespace

src/tasks/utils/cnf_installation/config_transformer.cr
```diff
@@ -0,0 +1,131 @@
...
+module CNFInstall
+  class ConfigTransformer
+    # Initialize with the original configuration hash
+    def initialize(@old_config : Hash)
+    end
+
+    # Perform the transformation to the new config format
+    def transform : Hash
+      transformed_config = {
+        "config_version" => "2.0", # New version identifier
```
kosstennbl, line 12: Making current config version a constant in CNFInstall module fells like a better solution

kosstennbl (changes requested):
Overall comment - feels a bit too repetitive, maybe consider adding some private methods to the class that would shorten the code.

svteb:
> Additionally, how about error handling? Like non-supported config keys on input? Maybe other error conditions?

I believe that this is out of scope of this ticket. These could all be considered validation. 

To my knowledge the v1 config does not have a validator that we could reuse to verify the input data, but considering that all the sample/example cnfs are currently (hopefully) deployable I do not find it necessary to verify the input data. Afterall they will all be replaced in step 3 of #2120 epic.

As for the output validation, that also probably does not fall in scope of this ticket as that should be included in #2129 which is the main parser for v2. The current flow of the program for v2 config should be something like this:

1. `./cnf-testsuite cnf_setup config.yml`
2. Execute parser from #2129  ~> Checks if the config is v1 or v2.
3. If v1, execute the transformer code which will return a transformed config back to the parser from #2129.
4. Parser then validates the input, etc.

As for error handling, the current format of `"key" => @old_config["key"]?.as_s` assures that if some key is not present its value will be stored as nil (which is inconsequential as the v2 class will also have nil values). Non-supported config keys are not recognized by the transformer because it looks only for the valid keys. In case of nested keys, I first check for existence of their parent. There should not be any conditions under which an error is given.

**Input validation:**
I simply don't find this necessary, do we really want to spend time on validating the v1 config if we intend to replace it in the next couple of weeks and the testsuite worked without it for some time now?

svteb:
> Can the architecture be made more future-proof towards possible future config versions?
> 
> * Identify source version of the confing on the input
> * Have possibility to tell which config version one needs to transform to (perhaps with hardcoded config version actually used by the testsuite as a constant somewhere)
> * Make an easy way for future introduction of config v3.0 and newer, with possibility to transform from 1.0 and 2.0

This is a valid point, I will attempt to generalize the process/add more private functions.

src/tasks/utils/cnf_installation/config_transformer.cr
svteb: Interim solution.

`"container_names" => @old_config["container_names"]?.as_a,`

src/tasks/utils/cnf_installation/config_transformer.cr
svteb: Removed and will not be stored.

src/tasks/utils/cnf_installation/config_transformer.cr
kosstennbl: My mistake, old config doesn't have additional structure for 5g parameters.

svteb:
**Short summary of what was attempted during implementation and the solution:**

I made multiple attempts at creating a generic solution that would allow for future extendibility, these have unfortunately hit some really hard roadblocks. The outline for a generic solution encompassed these properties:

1. Output of the transformer is easily parsed by the appropriate config class it is being converted to.
2. Future config formats (v2,3,...) would not require any changes to the main ConfigTransformer class, they would merely require a new appropriate rule set made up of transformations to be added.
3. The transformation rule sets would be easy to write (old key chain -> new key chain).
4. The transformer would resolve any reasonably expected/correct YAML configs.

**The approaches:**
**Rule sets**
The advanced approach would allow for rules like these to be defined:

```
[:container_names, :rolling_version_upgrade] -> [:common_parameters, :container_names, :rolling_version_upgrade]
[:helm_repo_name] -> [:deployments, :helm_charts, :helm_repo_name]
...
```
These would be easy to write, specifying where some prior key path should lead in the new config. The problems that come with this approach are numerous: 
1. No easy way to resolve keys that repeat. In the above example the container_names can have many rolling_version_upgrade keys inside of it as part of a YAML sequence. This could be resolved by adding an additional value to the rule, something like `key_repeats = true`. But that would bloat the rules quite a bit and could be hard to understand for future users.
2. Repeat keys must be transformed in the correct order, that is a sequence of container_names must first finish and only then can another sequence of container_names be started. This could only be resolved by detecting the type of container_names as an array and the correctly attempting to parse it which would require some type of recursive/stack approach. This was determined to be quite unreadable.
3. Extracting data from a class like this with a simple key chain requires quite a bit of hacking, such as generating getter functions for all parameters using a macro, additionally we run into return types like Class or Array(Class) which crystals typing system is not fond of.

A simplified approach is to simply hard-code what variable should be assigned where, i.e. `common_params["white_list_container_names"] = @old_config.white_list_container_names.to_s`. This does not look as pretty but the jump from specific solution to a generic one is too vast.


**Transformations**
1. Making use of YAML::Builder methods to construct a String.
This is supposedly the correct approach but it is exceedingly hard to implement as all possible types need to be resolved, this includes the types like Class/Array(Class). The key chains need to be order lexicographically because there is no possibility to move around the keys in String once its been constructed. Additionally there is a lot of prechecking necessary to determine which key chain should be constructed (to avoid empty keys), this gets harder with repeat keys/sequences. Also the YAML::Builder is pretty poorly documented and the solution that would come out of it would be very long, creating a brand new "reverse YAML parser" that is worthy of its own library module should not be within the scope of this ticket.
2. Class to class transformation.
This could be a relatively short solution but it would necessitate that the target class has initializers for passed values, this would make the target class quite bloated (as currently the expected initialize method is through a YAML file). Having a future requirement like this for the users is likely not what we want.
3. Creating a temporary Hash/YAML::Any structure on the go while adding data from the old config. While this is not a pretty solution it allows for some semblence of a generic approach. The ConfigTransformer itself would not need to be rewritten in case of future extension, there would only need to be a pretty hefty transformation class/function that is passed to it. While not the prettiest, it is easily more readable and allows for quick fixes. It also does not impose any requirements on other parts of the codebase.

As things stand the third option is the only one that does not impose on other parts of the codebase and is relatively readable/short. While it does not grant an easy way for future users to transform to a new config, the construction of new transformation rule set should not take too long.

svteb:
@martin-mat @kosstennbl @barmull ready for review.

src/tasks/utils/cnf_installation/transformer/transformation_base.cr
```diff
@@ -0,0 +1,67 @@
...
+    # functions at the end of its transform function.
+    class TransformationBase
+      @new_config : YAML::Any
+  
+      def initialize(@old_config : ConfigV1)
+        @new_config = YAML::Any.new({} of YAML::Any => YAML::Any)
+      end
+  
+      # Recursively converts the Hashes, Arrays and raw values to YAML::Any
+      private def hash_to_yaml_any(data : Hash | Array | String | Nil) : YAML::Any
```
kosstennbl, line 17: Probably can be renamed as it transforms not only hashes, but also arrays and strings.

src/tasks/utils/cnf_installation/transformer/transformation_base.cr
```diff
@@ -0,0 +1,67 @@
...
+          # Handle Hashes
+          converted_hash = Hash(YAML::Any, YAML::Any).new
+          data.each do |k, v|
+            converted_hash[YAML::Any.new(k)] = hash_to_yaml_any(v)
+          end
+          YAML::Any.new(converted_hash)
+        else
+          # Handle String or Nil cases
+          YAML::Any.new(data)
+        end
```
kosstennbl, line 34: Could be replaced by "case" statement.
case data
when Array
when Hash
else

src/tasks/utils/cnf_installation/transformer/transformation_base.cr
```diff
@@ -0,0 +1,67 @@
...
+          cleaned_array = Array(YAML::Any).new
+          array.each do |element|
+            cleaned_element = remove_nil_values(element)
+            cleaned_array << cleaned_element unless cleaned_element.nil?
+          end
+          return cleaned_array.empty? ? nil : YAML::Any.new(cleaned_array)
+        else
+          return data
+        end
+      end
```
kosstennbl, line 64: Whole function can be rewritten to a case:
Case data
when Nil
  nil
when Hash
  ...
when Array
  ...
else
  data

Also, specifically writing "return" statement wouldn't be needed in this solution.

src/tasks/utils/cnf_installation/transformer/v1_to_v2_transformation.cr
```diff
@@ -0,0 +1,115 @@
...
+          }]
+        elsif @old_config.helm_repository
+          deployments["helm_charts"] = [{
+            "name" => @old_config.release_name,
+            "helm_repo_name" => @old_config.helm_repository.not_nil!.name,
+            "helm_repo_url" => @old_config.helm_repository.not_nil!.repo_url,
+            "helm_chart_name" => @old_config.helm_chart,
+            "helm_values" => @old_config.helm_values,
+            "namespace" => @old_config.helm_install_namespace
+          }]
```
kosstennbl, line 79: "helm_repository" isn't a mandatory config parameter. It should transform if "helm_chart" parameter exists, and if "helm_repository" is also present - it should fill out "helm_repo_name" and "helm_repo_url"

src/tasks/utils/cnf_installation/transformer/v1_to_v2_transformation.cr
```diff
@@ -0,0 +1,115 @@
...
+  module Config
+    # Rules for configV1 to configV2 transformation
+    class V1ToV2Transformation < TransformationBase
+      def transform : YAML::Any
+        new_config_hash = {
+          "config_version" => "v2",
+          "common_parameters" => transform_common_parameters,
+          "dynamic_parameters" => transform_dynamic_parameters,
+          "deployments" => transform_deployments,
+          "five_g_parameters" => transform_five_g_parameters
```
kosstennbl, line 11: "five_g_parameters" are under common parameters, should be transformed there.

kosstennbl (changes requested):
Good work done on that feature, here's few notes from review:

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,110 @@
...
+    class ConfigBase
+      include YAML::Serializable
+      include YAML::Serializable::Strict
+    end
+
+    enum ConfigVersion
+      First = 1
+      Second = 2
+      Latest = Second
+    end
```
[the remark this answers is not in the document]
svteb: `enum` only supports integeres.

src/tasks/utils/cnf_installation/transformer/transformation_base.cr
svteb: Unfortunately due to crystals terrible type detection this cannot be a case statement, I've attempted it previously. The `is_a` function does some deeper detection that finds out what the underlying type of `YAML::Any` is.

src/tasks/utils/cnf_installation/transformer/transformation_base.cr
svteb: Similarly to above, this cannot be a case statement, but I will think about removing the `return` as it also strikes me as unnecessary.

kosstennbl:
Re-created from different branch as #2147

Issue #2136, opened 2024-08-16
[BUG] Image size spec test using protected repo is not working


martin-mat:
**Describe the bug**
When executing spec test for image size with protected repo, it fails on helm install:
![image](https://github.com/user-attachments/assets/b0a56b63-5859-42be-ab20-dfcb3cd94f8e)

https://github.com/cnti-testcatalog/testsuite/actions/runs/10405256412/job/28844883137

**To Reproduce**
Run reasonable_image_size spec tests on the main branch (not from forked repo)

svteb:
Bit of a duplicate of #2081

PR #2137, opened 2024-08-16
quick fix: remove protected image repo spec test

martin-mat:
After #2114 was merged, an issue was revealed
with image size - protected repo spec test.
It was discovered only after merging.
Until this is fixed, disabling this variant
of the image size spec test.

Refs: #2114 #2136


svteb (approved):
lgtm, maybe correct the identation a little.

kosstennbl (approved):
LGTM, temporary disabled, later should be reviewed to be deleted or fixed

Smitholi67 (approved):
lgtm

Issue #2138, opened 2024-08-19
[Feature] CNF Installation (2.2), New config format generation


kosstennbl:
Testsuite has useful generate_config functionality that allows users to generate basic config for their CNF without knowledge of all the parameters.

With update of config format - we need a new way of config generation. It should be able to create cnf-testsuite.yml for 3 basic deployment types containing all the required parameters (Deployment name, type-specific parameters).

New spec tests should also be added.

kosstennbl:
Done #2148

PR #2142, opened 2024-09-04
config-generator: CNF Installation (2.2) Add generator for v2 config

kosstennbl:
## Description
Create new config generation task, that would generate v2 config.
This task should be more interactable and friendly to new testsuite users.

## Issues:
Refs: #2138


kosstennbl:
Can be merged only after #2129

svteb:
Perhaps change the title so it conforms to the other epic tasks so far?

src/tasks/cnf_setup.cr
```diff
@@ -99,3 +104,69 @@ end
...
+
+  loop do
+    puts "Select deployment type:"
+    puts "1. Helm Chart"
+    puts "2. Helm Directory"
+    puts "3. Manifest Directory"
+    puts "4. Finish and save configuration"
+    print "Enter your choice (1-4): "
+    
+    choice = gets.try(&.strip)
```
svteb, line 126: Why does this not use the `prompt` function?

kosstennbl: Fixed

svteb (approved):
I presume that the changes from #2129 that have been copied into this pull request will be removed before merging. Nevertheless, works as a good quick-start implementation, lgtm.

kosstennbl:
Re-created from different branch as #2148

PR #2147, opened 2024-09-16
config: CNF installation (2.1) Transformer for old version of config

svteb:
## IMPORTANT
To speed up the development of the #2120 epic we are currently experimenting with a different approach to pull requests. This pull request has been created anew from a different branch, if you want to check the already existing discussion around it, view the old [#2133](https://github.com/cnti-testcatalog/testsuite/pull/2133).

## Description
The transformer transforms the old config according to the structure proposed in #2129.

**Notes**
1. To allow for some input validation, the v1 config was reimplemented as class in `config_v1.cr`.
2. Certain arguments are found in cnf-testsuite.yml files spread across the samples and examples that are unused anywhere in the code, these are still parsed but ignored to avoid too much manual intervention in later stages.
3. A new task has been added to allow for transformation, users can execute it by calling this command:

`./cnf-testsuite transform_config OLD_FILE_PATH NEW_FILE_PATH`

4. The `ConfigTransformer` class found in `config_transformer.cr` automatically detects the version of the old config and transforms it to the latest version (currently only v1 -> v2), but allows for future extendibility.
5. Upon determining the version of the old config, the main `transform` function is called which does the actual transformation in code (through use of Hashes and YAML::Any).
6. The `transform` function uses the appropriate transformation rules (`V1ToV2Transformation class` in `v1_to_v2_transformation.cr` file)
7. To make future extension of transformation rules easier, the function `process_data` has been added (`TransformationBase class` in `transformation.cr` file), this function converts the underlying hashes and arrays to the `YAML::Any` type and removes any nil branches so they don't clutter the final output.
8. After the transformation is done the result can be printed to either String or dumped to a file through the `serialize_to_string` and `serialize_to_file` functions.
9. Most edge cases have been resolved through the use of exceptions and their catching, giving users a readable output in case of any errors.

## Issues:
Refs:  #2130

## How has this been tested:
 - [x] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] My change requires a change to the documentation.
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/CNF_TESTSUITE_YML_USAGE.md
+++ b/CNF_TESTSUITE_YML_USAGE.md
@@ -43,6 +43,14 @@ Prereqs: You must have kubernetes cluster, curl, and helm 3.1.1 or greater on yo
 - Generate a cnf-testsuite.yml based on a directory of manifest files:  `./cnf-testsuite generate_config config-src=<your-manifest-directory> output-file=./cnf-testsuite.yml`
 - Inspect the cnf-testsuite.yml file for accuracy
 
+### Updating config from older versions
+
+New releases may change the format of cnf-testsuite.yml. To update your older configs automatically to the latest version, use the `update_config` task.
+
+```
+./cnf-testsuite update_config input_config=OLD_CONFIG_PATH output_config=NEW_CONFIG_PATH
+```
+
 ### Keys and Values
 
 #### allowlist_helm_chart_container_names
--- a/src/tasks/update_config.cr
+++ b/src/tasks/update_config.cr
@@ -1,0 +1,45 @@
+require "sam"
+require "totem"
+require "colorize"
+require "./utils/cnf_installation/config_updater/config_updater"
+require "./utils/cnf_installation/config"
+
+desc "Updates an old configuration file to the latest version and saves it to the specified location"
+task "update_config" do |_, args|
+  # Ensure both arguments are provided
+  if !((args.named.keys.includes? "input_config") && (args.named.keys.includes? "output_config"))
+    stdout_warning "Usage: update_config input_config=OLD_CONFIG_PATH output_config=NEW_CONFIG_PATH"
+    exit(0)
+  end
+
+  input_config = args.named["input_config"].as(String)
+  output_config = args.named["output_config"].as(String)
+
+  # Check if the input config file exists
+  unless File.exists?(input_config)
+    stdout_failure "The input config file '#{input_config}' does not exist."
+    exit(1)
+  end
+
+  begin
+    raw_input_config = File.read(input_config)
+
+    # Verify that config is not the latest version
+    if CNFInstall::Config.config_version_is_latest?(raw_input_config)
+      stdout_warning "Input config is the latest version."
+      exit(0)
+    end
+
+    # Initialize the ConfigUpdater
+    updater = CNFInstall::Config::ConfigUpdater.new(raw_input_config)
+    updater.transform
+
+    # Serialize the updated config to the new file
+    updater.serialize_to_file(output_config)
+
+    stdout_success "Configuration successfully updated and saved to '#{output_config}'."
+  rescue ex : CNFInstall::Config::UnsupportedConfigVersionError
+    stdout_failure ex.message
+    exit(1)
+  end
+end
--- a/src/tasks/utils/cnf_installation/config.cr
+++ b/src/tasks/utils/cnf_installation/config.cr
@@ -28,5 +28,26 @@ module CNFInstall
 
       config
     end
+
+    # Detects the config version.
+    def self.detect_version(tmp_content : String) : ConfigVersion
+      yaml_content = YAML.parse(tmp_content).as_h
+      version_value = yaml_content["config_version"]?.try(&.to_s)
+
+      if version_value
+        begin
+          ConfigVersion.parse(version_value.upcase)
+        rescue ex : ArgumentError
+          raise UnsupportedConfigVersionError.new(version_value)
+        end
+      else
+        # Default to V1 if no version is specified
+        ConfigVersion::V1
+      end
+    end
+
+    def self.config_version_is_latest?(tmp_content : String) : Bool
+      detect_version(tmp_content) == ConfigVersion::Latest
+    end
   end
 end
--- a/src/tasks/utils/cnf_installation/config_updater/config_updater.cr
+++ b/src/tasks/utils/cnf_installation/config_updater/config_updater.cr
@@ -1,0 +1,76 @@
+require "yaml"
+
+module CNFInstall
+  module Config
+    class ConfigUpdater
+      @output_config : YAML::Any
+      @input_config : ConfigBase
+      @version : ConfigVersion
+
+      # This approach could be extended in future by making use of abstract classes,
+      # which would remove the need for hashes. 
+      # Define transformation rules at the top of the class
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      VERSION_TRANSFORMATIONS = {
+        ConfigVersion::V1 => ->(input_config : ConfigBase) { V1ToV2Transformation.new(input_config.as(ConfigV1)).transform }
+      }
+
+      # Define parsing rules at the top of the class
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      VERSION_PARSERS = {
+        ConfigVersion::V1 => ->(raw_input_config : String) { ConfigV1.from_yaml(raw_input_config) }
+      }
+
+      def initialize(raw_input_config : String)
+        # Automatic version detection to streamline the transformation
+        @version = CNFInstall::Config.detect_version(raw_input_config)
+        @output_config = YAML::Any.new({} of YAML::Any => YAML::Any)
+        @input_config = parse_input_config(raw_input_config)
+      end
+
+      # Serialize the updated config to a string.
+      def serialize_to_string : String
+        YAML.dump(@output_config)
+      end
+
+      # Serialize the updated config to a file and return the file path.
+      def serialize_to_file(file_path : String) : String
+        File.write(file_path, serialize_to_string)
+        file_path
+      end
+
+      # Parses the config to the correct class.
+      # Uses the VERSION_PARSERS hash.
+      private def parse_input_config(raw_input_config : String) : ConfigBase
+        parser = VERSION_PARSERS[@version]
+        if parser
+          begin
+            parser.call(raw_input_config)
+          rescue ex : YAML::ParseException
+            stdout_failure "Failed to parse config: #{ex.message}."
+            exit(1)
+          end
+        else
+          raise UnsupportedConfigVersionError.new(@version)
+        end
+      end
+
+      # Performs the transformation from Vx to Vy.
+      # Uses the VERSION_TRANSFORMATIONS hash.
+      def transform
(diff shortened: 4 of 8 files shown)


CNF_TESTSUITE_YML_USAGE.md
```diff
@@ -43,6 +43,14 @@ Prereqs: You must have kubernetes cluster, curl, and helm 3.1.1 or greater on yo
 - Generate a cnf-testsuite.yml based on a directory of manifest files:  `./cnf-testsuite generate_config config-src=<your-manifest-directory> output-file=./cnf-testsuite.yml`
 - Inspect the cnf-testsuite.yml file for accuracy
 
+### Transformer for prior config versions
```
martin-mat, line 46: I'd suggest to use more understandable
"Updating config from older versions"

also consider replacing "transformer" to "updater" across this change as this in my opinion better describes its functionality.

CNF_TESTSUITE_YML_USAGE.md
```diff
@@ -43,6 +43,14 @@ Prereqs: You must have kubernetes cluster, curl, and helm 3.1.1 or greater on yo
...
+
+New releases may change the format of cnf-testsuite.yml, to quickly update your older configs to latest version use the `transform_config` task.
```
martin-mat, line 48: I'd suggest:
`New releases may change the format of cnf-testsuite.yml. To update your older configs automatically to the latest version, use the 'update_config' task.`

Then: how is the user supposed to know whether this action is needed? Is there any deprecation warning message displayed in  case an old version of the config is used? I think it should be. Such deprecation message should be then mentioned here in the docs.

CNF_TESTSUITE_YML_USAGE.md
```diff
@@ -43,6 +43,14 @@ Prereqs: You must have kubernetes cluster, curl, and helm 3.1.1 or greater on yo
...
+
+```
+./cnf-testsuite transform_config old_config_path=OLD_CONFIG_PATH new_config_path=NEW_CONFIG_PATH
```
martin-mat, line 51: update_config
and I would also suggest to change
"old_config_path", "new_config path"
to
"input_config" and "output_config"

src/tasks/transform_config.cr
```diff
@@ -0,0 +1,37 @@
+require "sam"
+require "totem"
+require "colorize"
+require "./utils/cnf_installation/transformer/config_transformer"
+
+desc "Converts an old configuration file to the latest version and saves it to the specified location"
```
martin-mat, line 6: `Updates an old configuration file to the latest version and saves it to the specified location`

src/tasks/transform_config.cr
```diff
@@ -0,0 +1,37 @@
...
+task "transform_config" do |_, args|
```
martin-mat, line 7: update_config

src/tasks/transform_config.cr
```diff
@@ -0,0 +1,37 @@
...
+    # Initialize the ConfigTransformer
+    transformer = CNFInstall::Config::ConfigTransformer.new(old_config_path)
+    transformer.transform
+
+    # Serialize the transformed config to the new file
+    transformer.serialize_to_file(new_config_path)
+
+    stdout_success "Configuration successfully transformed and saved to '#{new_config_path}'."
+  rescue ex : CNFInstall::Config::UnsupportedConfigVersionError
+    stdout_failure ex.message
```
martin-mat, line 33: Not sure what kind of message is printed out here.
If this is an user-targeted error: a clear user-understandable description should be printed out. Is it?
If this is a system-level error: I think then it should be logged to the log channel as a warning/error depending on severity.

Additionally, should the task exit with an error code?

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,97 @@
...
+          case @version
+          when ConfigVersion::V1
+            ConfigV1.from_yaml(tmp_content)
+          else
+            raise UnsupportedConfigVersionError.new(@version)
+          end
+        rescue ex : YAML::ParseException
+          # This is usually raised in case of unexpected YAML keys.
+          stdout_failure "Failed to parse config: #{ex.message}."
+          stdout_failure "Please check your YAML fields for correctness."
```
martin-mat, line 74: Avoid presonal pronouns

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,97 @@
...
+        begin
+          case @version
+          when ConfigVersion::V1
+            ConfigV1.from_yaml(tmp_content)
+          else
+            raise UnsupportedConfigVersionError.new(@version)
+          end
+        rescue ex : YAML::ParseException
+          # This is usually raised in case of unexpected YAML keys.
+          stdout_failure "Failed to parse config: #{ex.message}."
```
martin-mat, line 73: operator understandable message -> stdout
developer level message -> log ERROR

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,97 @@
...
+        else
+          ConfigVersion::V1
+        end
+      end
+
+      # Parses the config to the correct class.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      private def parse_old_config(tmp_content : String) : ConfigBase
+        begin
+          case @version
```
martin-mat, line 65: better to use hash structures than case.

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,97 @@
...
+          stdout_failure "Failed to parse config: #{ex.message}."
+          stdout_failure "Please check your YAML fields for correctness."
+          exit(1)
+        end
+      end
+
+      # Performs the transformation from V1 to V2.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      def transform
+        case @version
```
martin-mat, line 82: better to use hash structures than case

martin-mat (changes requested):

CNF_TESTSUITE_YML_USAGE.md
svteb: This is currently out of scope of the ticket, @kosstennbl will implement a check in one of his components where the main parsing happens (in case the version is outdated we will run the ConfigUpdater automatically/tell the user to run it themself), although we have agreed on me providing some method that detects if config is the latest version.

src/tasks/transform_config.cr
svteb: image

This is mostly a programmer-targeted output, a code-induced check if you will. In case of future extension the `case` statements in `ConfigUpdater` methods need to be upgraded. Since there are quite a few, the programmer might miss one of them and this is the way of telling him that he did. The only other way this could be triggered is if the user manually set his config_version field to something unexpected (like v3 right now), in which case it still serves the purpose.

src/tasks/transform_config.cr
```diff
@@ -0,0 +1,37 @@
...
+  rescue ex : Exception
+    stdout_failure "Unexpected error: #{ex.class}"
```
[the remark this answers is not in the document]
svteb: The ex.class is an artifact of my testing and should be ex.message here.

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
```diff
@@ -0,0 +1,97 @@
...
+
+      # Detects the config version.
+      # REQUIRES FUTURE EXTENSION in case of new config format.
+      private def detect_version(yaml_content : Hash(YAML::Any, YAML::Any)) : ConfigVersion
+        version_value = yaml_content["config_version"]?.try(&.to_s)
+
+        if version_value
+          begin
+            ConfigVersion.parse(version_value.upcase)
+          rescue ex : Exception
```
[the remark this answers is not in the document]
svteb: While I understand this, this is such a miniscule block that there really isnt any reason to not just rescue the Exception, but I'll change it to `ArgumentError`.

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
svteb: Yes, this is message for operators, the ex.message will say something along the lines of "Could not parse x field".

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
svteb: I will rewrite it, but what is the purpose? We don't exactly shorten the code, readability remains the exact same if not slightly worse, and the speed difference is nil.

CNF_TESTSUITE_YML_USAGE.md
svteb: I disagree with changing transformer to updater throughout the code, we are changing the entire schema of the config (transforming it), an updater would be an appropriate name if we simply updated the values of the config. There is also the TransformationBase class, changing that to UpdateBase just sounds silly to me. I would be more inclined to call it an Upgrader, but I still think that Transformer is the correct choice.

CNF_TESTSUITE_YML_USAGE.md
svteb: Would this be alright?

`
./cnf-testsuite update_config input_config=OLD_CONFIG_PATH output_config=NEW_CONFIG_PATH
`

I am afraid that doing input_config=INPUT_CONFIG, output_config=OUTPUT_CONFIG would not be descriptive enough.

CNF_TESTSUITE_YML_USAGE.md
svteb: After some thought, we could maybe do a mixed approach, because TransformationBase -> UpdaterBase just does not sit well with me, but other parts would get changed to updater.

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
martin-mat: Code maintainability. For cases when:
- adding a new config version -> add appropriate records to a hash somewhere in the config structure in the beginning of a file, no need to go through all the code and search for all cases where version is considered.
- adding new call dependent on config version somewhere in the code. Then just use a function call from a hash structure instead of constructing additional case command with listing (in future) multiple config versions.

if in doubts we can discuss f2f.

src/tasks/utils/cnf_installation/transformer/config_transformer.cr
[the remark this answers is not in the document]
martin-mat: Special cases don't jsutify bad habbits :-)

CNF_TESTSUITE_YML_USAGE.md
martin-mat: Yeah, consider what makes best sense from code architecture/meaning point of view.

kosstennbl (approved):
LGTM

martin-mat (approved):
lgtm

Smitholi67 (approved):
LGTM

Issue #2157, opened 2024-09-26
[MAINTENANCE] Remove Ameba from shard dependencies


kosstennbl:
Ameba is a crystal-lang static linting tool, which can be ran in different ways. https://github.com/crystal-ameba/ameba

Currently we experience issues with upgrading crystal version due to incompatible versions of ameba and crystal.
And even through Ameba is mentioned in dependencies of testsuite, it isn't being used in any way for CI or otherwise.

Proposal is to remove Ameba from project dependencies, and then, in future issues, update crystal version and later, add Ameba to CI via building it from source.

No other projects under testsuite use Ameba, it should be safe to only remove it from shard.yml and shard.lock here.

Issue #2159, opened 2024-09-26
[Feature] Add Ameba to CI runs


kosstennbl:
If Ameba would be removed from dependencies and crystal-version would be updated (#2157, #2158), we can re-add Ameba to testsuite, but not in form of shard dependency, but rather as a tool in our CI

Proposal is to add a new workflow to GitHub actions, which would build ameba from sources and then run static lint on a repo.
Together with that, all issues would have to be fixed (1500+ LOC), but most of them are very simple and can be auto-corrected.

Instructions for installation and usage:
https://github.com/crystal-ameba/ameba

PR #2160, opened 2024-09-27
ameba: Remove ameba from shard dependencies

kosstennbl:
## Description
Ameba is a linting tool, which isn't being used in our CI pipelines, and causes complications with crystal version update. 

Ameba dependency should be removed from cnf-testcatalog shards and usage guidelines should point to more independent usage via installing the linter from source.

## Issues:
Refs: #2157


martin-mat (approved):
lgtm as addressing immediate issues with ameba.
Target solution to be addressed as a part of #2159

Smitholi67 (approved):
LGTM.  Perhaps we should have the same approach to our other repos even though there is not as much activity in those at the moment.  Nothing urgent but something good if we are consistent.

svteb (approved):
lgtm

Issue #2223, opened 2025-02-24
[BUG] sig_term test is unstable in github actions


martin-mat:
sig_term test often fails in gh actions. Examples:

https://github.com/cnti-testcatalog/testsuite/actions/runs/13498980714/job/37712588374
https://github.com/cnti-testcatalog/testsuite/actions/runs/13108987316/job/36574062616

![Image](https://github.com/user-attachments/assets/a908abc0-bfd7-4c20-99bb-80d59764f9e3)

PR #2235, opened 2025-03-11
Fix: Harden sig_term_handled test

svteb:
## Description
- Prior implementation of the test did not correctly handle `strace` failures, resulting in the task being fragile and failing on ephemeral/untracable processes.
- Above issues were resolved by adding correct error handling on expected errors.
- The code structure was also flattened for better readability.

## Issues:
Refs: #2223

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/workload/microservice.cr
+++ b/src/tasks/workload/microservice.cr
@@ -21,6 +21,13 @@ task "microservice", ["reasonable_image_size", "reasonable_startup_time", "singl
 end
 
 REASONABLE_STARTUP_BUFFER = 10.0
+STRACE_WAIT_BUFFER = 3.0
+
+enum StraceAttachResult
+  Attached
+  NotPermitted
+  NoSuchProcess
+end
 
 desc "To check if the CNF has multiple microservices that share a database"
 task "shared_database", ["install_cluster_tools"] do |t, args|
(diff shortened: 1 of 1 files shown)


src/tasks/workload/microservice.cr
```diff
@@ -459,16 +466,35 @@ task "zombie_handled" do |t, args|
   end
 end
 
+# Attach strace to a PID in background
+def attach_strace(pid : String, node : JSON::Any)
+  log_path = "/tmp/#{pid}-strace"
+  cmd = "strace -p #{pid} -e 'trace=!all' 2>&1 | tee #{log_path}"
+  ClusterTools.exec_by_node_bg(cmd, node)
+
+  sleep 1  # give strace time to report immediate errors
```
rafal-lal, line 475: This is checking initial set of logs that will be in the file always after starting right? You are checking for well known strings listed below like "Operation not permitted" or "No such process" and other. What about instead of waiting 1s which is not reliable and can also be flaky, wait until one of the strings you check for will show in the logs. That is of course based on assumption that there is a set of these initial messages that will provide us this info. Give it some timeout to be sure as well. Then we will have the certainty, right now after 1 sec there is unlikely possibility that strace havent reported anything yet or any other edge case that will give us unconsistent, hard to detect bug.

rafal-lal (changes requested):
LGTM except usage of sleep. What about error handling in kubectl calls? There were none previously but we have a Issue about adding those.

src/tasks/workload/microservice.cr
svteb: Yeah, that sounds much better, I'll rework it with some actual hard timeout limits, first checking if the file is created and and then interpret the content.

As for handling kubectl calls, I sort of forgot that we have that now :D .

src/tasks/workload/microservice.cr
```diff
@@ -459,16 +466,35 @@ task "zombie_handled" do |t, args|
   end
 end
 
+# Attach strace to a PID in background
+def attach_strace(pid : String, node : JSON::Any)
+  log_path = "/tmp/#{pid}-strace"
+  cmd = "strace -p #{pid} -e 'trace=!all' 2>&1 | tee #{log_path}"
+  ClusterTools.exec_by_node_bg(cmd, node)
```
rafal-lal, line 473: `strace` starts in the background so it needs to be somehow stopped after the test. I don't think it stops by itself but I will leave the investigation to you.

rafal-lal: Bad thing is this is started by some ClusterTool method, we might need some bigger reworks to make this work in sensible way so if there is no easy way I think we can skip it for now and make an issue about it. ClusterTools will be updated soon anyway.

rafal-lal: Sounds good to me, as stated we may look for some in-code solution but for now this is perfect.

rafal-lal (approved):
lgtm

martin-mat (approved):
lgtm

collivier (approved):

Issue #2199, opened 2025-01-07
[Feature] Refactor Logging System: Reduce Noise & Add more informational logs


svteb:
**Is your feature request related to a problem? Please describe.**
There are currently too many logs that do not provide any useful information and merely cause noise. Certain logs are incorrectly categorized (INFO should be DEBUG and vice versa). Additionally it is hard to determine the cause of failure in many tests since we do not log the state of the cluster after failure (test resources, deployments, etc.)

**Describe the solution you'd like**

**Noise:**
- [ ] Consider shortening the log header itself, the current format is `I, [2025-01-03 07:23:11 +00:00 #189258]  INFO -- `, it is quite apparent that we do not need to specify `I` and `INFO` as they are the same. The UTC offset and process PID also seems useless. 
- [ ] Utilize the `Log.for()` to scope the logs more appropriately (it takes considerable time to figure out where a log is coming from in the first place)
- [ ] The `cnf_install` (formerly `cnf_setup`) execution prints 4 lines of logs every second. Should be drastically reduced.
- [ ] Miscellaneous logs with no apparent meaning should be removed/refactored.
- [ ] Certain logs print out entire network statuses, manifests, process information and the like. These can be significantly reduced by only printing out information relevant to the test.
- [ ] Certain logs could be batched together into a single array (there are various for loops where an IP address/network status are printed on separate lines when looking for a "match").
- [ ] In various parts of the testsuite a `Unable to use a TTY - input is not a terminal or the right kind of file` error is printed out, determine the cause and remove it.
- [ ] Determine case by case which logs really belong to INFO and DEBUG categories.
- [ ] Format the printed logs in a logical manner (use `\n` in case of multiline outputs, e.g. so that manifests are properly formatted and the first line is not offset).

- Most of the noise comes from the libraries (shards) that testsuite makes use of.

**Information:**
- [ ] After failure the cluster resources should be printed out (figuring this out could prove to be more difficult than apparent).
- [ ] Attempt to make the INFO logs more "humanly" readable. 

**Describe alternatives you've considered**
- As the testsuite slowly transitions into a more OOP model, we could consider implementing an inheritable class to ease log scoping for classes (research more into `Log.for` and `Log.builder`).

**Additional context**
This ticket will be updated as new problems are found, please comment if you have any other ideas/findings. As these changes do not only span tens of files but also numerous libraries, the changes should not come in a single pull request, but be split, to allow a sensible review process.

PR #2252, opened 2025-04-01
Refactor 'setup' part of the testsuite

rafal-lal:
## Description
Refactor 'setup' part of the testsuite with focus on logs, removing unused stuff and cleaning the code. Subsequent PRs will be added for each tool that is being set up to make changes reviewable.

## Issues:
Refs: #2199

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [x] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


src/tasks/utils/utils.cr
```diff
@@ -380,3 +375,20 @@ def self.find(directory : String, wildcard : String)
 
   found_files
 end
+
+def download(url : String, output_path : String,
```
martin-mat, line 379: That looks too generic function name. Maybe "download_file" or "http_download"?

src/tasks/utils/utils.cr
```diff
@@ -368,7 +363,7 @@ def with_kubeconfig(kube_config : String, &)
   result
 end
 
-def self.find(directory : String, wildcard : String)
+def find(directory : String, wildcard : String)
```
martin-mat, line 366: "find" is too generic. Maybe "find_files"?

martin-mat (changes requested):

svteb (changes requested):
I disagree with the removal of FluentD flavor, while it might not be used in the code explicitly, if you look at the implementation of `routed_logs`, we search for common known flavors of fluent and this is one of them. This approach isn't the best, but removing it at this time is not appropriate. I also don't understand the reason for the observability spec test removal.

Additionally, what about `src/tasks/prereqs`? Did you consider some changes with it/maybe including it in the setup namespace or moving it elsewhere (just looks like a good time to do something with it).

src/tasks/utils/utils.cr
svteb, line 379: I am wondering if we should create a custom download error exception for better handling, something like:

```
class DownloadError < Exception
  def initialize(message : String)
    super(message)
  end
end
```

then we could extend this function to include the expected HTTP::Client.get exceptions. This might be a little overkill but it would allow for more fidelity when rescuing.

```
  rescue ex : Socket::Error
    raise DownloadError.new("Network issue during download", ex)
  rescue ex : IO::TimeoutError
    raise DownloadError.new("Timeout during download", ex)
  rescue ex : OpenSSL::SSL::Error
    raise DownloadError.new("SSL error during download", ex)
  rescue ex : HTTP::Client::Error
    raise DownloadError.new("HTTP client error", ex)
```

We can discuss this.

spec/cluster_setup_spec.cr
```diff
@@ -4,12 +4,13 @@ require "../src/tasks/utils/utils.cr"
 
 describe "Cluster Setup" do
   it "'install_cluster_tools' should give a message if namespace does not exist", tags: ["cluster_setup"]  do
-    KubectlClient::Delete.resource("namespace", ClusterTools.namespace)
-    result = ShellCmd.run_testsuite("install_cluster_tools")
+    begin KubectlClient::Delete.resource("namespace", ClusterTools.namespace) rescue Exception end
```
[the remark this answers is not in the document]
rafal-lal: Done.

src/tasks/utils/utils.cr
rafal-lal: Sounds good.

src/tasks/utils/utils.cr
rafal-lal: agree.

spec/utils/utils_spec.cr
```diff
@@ -202,4 +203,104 @@ describe "Utils" do
...
+      when "/redirect"
+        response.status_code = 302
+        response.headers["Location"] = "http://#{ip}:#{port}/success"
+        response.print("redirecting to /success")
+      when "/error"
+        response.status_code = 404
+        response.print("not found")
+      when "/auth"
+        auth_header = request.headers["Authorization"]
+        if auth_header && auth_header.starts_with?("Bearer ")
```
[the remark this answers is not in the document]
rafal-lal: Yes and no, `Bearer` would generally work but edge case in which token is in bad format (misspelled) `Bearertokenwithoutspace`. This could happen when typing token manually or pasting. Normally I would ignore, but if it means just adding whitespace in one line I decided to go with it.

src/tasks/setup/cluster_api_setup.cr
```diff
@@ -0,0 +1,85 @@
...
+
+    clusterctl = Path["~/.cluster-api"].expand(home: true)
+
+    FileUtils.mkdir_p("#{clusterctl}")
+
+    File.write("#{clusterctl}/clusterctl.yaml", "CLUSTER_TOPOLOGY: \"true\"")
+
+    cluster_init_cmd = "clusterctl init --infrastructure docker"
+    stdout = IO::Memory.new
+    Process.run(cluster_init_cmd, shell: true, output: stdout, error: stdout)
```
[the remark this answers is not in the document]
rafal-lal: Fixed in subsequent cluster_api PR so ignoring it here.

src/tasks/setup/cluster_api_setup.cr
```diff
@@ -0,0 +1,85 @@
...
+    create_cluster_file = "#{current_dir}/capi.yaml"
+
+    create_cluster_cmd = "clusterctl generate cluster capi-quickstart   --kubernetes-version v1.24.0   --control-plane-machine-count=3 --worker-machine-count=3  --flavor development > #{create_cluster_file} "
+
+    Process.run(
+      create_cluster_cmd,
+      shell: true,
+      output: create_cluster_stdout = IO::Memory.new,
+      error: create_cluster_stderr = IO::Memory.new
+    )
```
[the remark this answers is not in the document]
rafal-lal: Same as above.

src/tasks/utils/utils.cr
rafal-lal: IMO this is bit too much and just exception rescuing might be good enough. I'm not saying your suggestion isnt valid, but how many different errors to check are there? And lets think about outcome of rescuing, will it change based on exception rescued or will it be logged the same way.

rafal-lal:
> I disagree with the removal of FluentD flavor, while it might not be used in the code explicitly, if you look at the implementation of `routed_logs`, we search for common known flavors of fluent and this is one of them. This approach isn't the best, but removing it at this time is not appropriate. I also don't understand the reason for the observability spec test removal.
> 
> Additionally, what about `src/tasks/prereqs`? Did you consider some changes with it/maybe including it in the setup namespace or moving it elsewhere (just looks like a good time to do something with it).

1. Let me investigate and fix that.
2. Lets discuss. I think prereqs belong in setup as well.

martin-mat (approved):
lgtm

svteb (approved):
lgtm

collivier (approved):
good work

PR #2260, opened 2025-04-08
Delete EmbeddedFileManager module

### LINKED ISSUE #2258: [MAINTENANCE] EmbeddedFileManager not being useful
**Describe the software update**
`EmbeddedFileManager` module which looks like this

Image

is not really doing anything useful. Files stored in `./embedded_files` directory are `cat`ted, then encoded with base64 cli tool and then decoded with base64 crystal-module. I think this was implemented this way as convenient way to store strings as CONSTs. Thing is that this is not really useful as those CONSTs are then saved to files again before applying. Example:

Image

Also couple of them are currently unused.

My proposal is to get rid of the overhead of this module completely and introduce one CONST with directory where files are stored and just use them directly.

rafal-lal:
## Description
This module job was to store manifests as constants inside memory but it
is not necessary as the constants content were then written to files
again before applying by kubectl. So the only thing it in fact did was
creating middleware layer that read files from one directory and save
them later in different, hence not giving any value to codebase but
rather providing code to maintain.

## Issues:
Refs: #2258 

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [x] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


martin-mat:
Closing this one because the premise conflicts with the current distribution model: the suite ships as a standalone static binary (see `release.yml`), and `EmbeddedFileManager` compile-time-embeds `embedded_files/*` precisely so that a downloaded binary works outside a source checkout. Replacing the embedding with runtime reads of `./embedded_files` would break `points.yml`, the OPA policies, fluent values, and the chaos manifests for anyone running the released binary. Main still fully relies on the embedding today (`src/tasks/constants.cr`, `points.cr`) and it's documented as the current design.

There's also a regression in the tip commit: `src/tasks/utils/ueransim.cr` renders crinja values into `gnb-ues-values.yaml` but then helm-installs with the un-templated `ue.yaml`, silently discarding the rendered config.

If the module bothers anyone, pruning unused macros — or a genuinely different design like runtime self-extraction — would be the way to go, but deleting the embedding isn't viable while we ship a single static binary.

Issue #2194, opened 2024-12-16
[BUG] Cleanup not waiting for resources to remove


martin-mat:
Cleanup does not wait for removing reources. This is inconsistent with how setup works, waiting for resources to be created.

collivier:
Agree. I did add a 60s sleep between to 2 cnf_testsuite calls (via Functest Kubernetes)

PR #2263, opened 2025-04-23
Feat: Add deletion timeout to cnf_uninstall

svteb:
## Description
This change adds failure checking and timeouts into the `cnf_uninstall` task, so that it mirrors the behavior of task `cnf_install` (adding uninstall timeout, making sure bad uninstall does not cause an exception, nicer user output in the terminal).

The meat of the change is in functions `wait_for_deployment_uninstallation` (which has been split from the original `wait_for_deployment_resources`) and `KubectlClient::Wait.resource_wait_for_uninstall`. The core issue was the method of detecting that something went wrong during uninstallation (imagine a stuck finalizer). The `kubectl delete` and `helm uninstall` commands do provide a `--wait` and `--timeout` flag that were attempted to be used, but this approach would not give us the fidelity of manual verification. The manual verification checks what resources have not been deleted and explicitly prints their failure to user. Since we implement manual verification with a set timeout value the `--wait` flag of `kubectl delete` and `helm uninstall` has been set to `false` by default.

As for the manual implementation in `KubectlClient::Wait.resource_wait_for_uninstall`, it checks for the existence of resource, but this is not sufficient by itself - in various cases the core deployment may have been deleted but its pods may remain (i.e. stuck finalizer) which is why we also need to pass labels, to look for pods in a secondary check.

Other big changes are relevant to the spec tests, to correctly verify that both manifest and helm uninstallation failures are detected two new spec tests were introduced, along with sample-cnfs (verification through stuck finalizers). To delete these stuck finalizers a new implementation of `kubectl patch` has been added to `KubectlClient::Utils.patch`. Another thing that I noticed while making this change was how badly the utility functions in `spec_helper.cr` were implemented (timeout was not functioning), which is why I did a minor change to the `ShellCmd.cnf_uninstall` function (full correction should be done when resolving #2280).

Other changes across the code simply correct function signatures.

## Issues:
Refs: #2194 

## How has this been tested:
 - [x] Added integration testing to cover
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [x] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/modules/helm/helm.cr
+++ b/src/modules/helm/helm.cr
@@ -278,13 +278,15 @@ module Helm
     ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
   end
 
-  def self.uninstall(release_name, namespace = nil) : CMDResult
+  def self.uninstall(release_name : String, namespace : String? = nil, wait : Bool = false) : CMDResult
     logger = Log.for("uninstall")
     logger.info { "Uninstalling helm chart: #{release_name}" }
 
     helm = BinarySingleton.helm
     cmd = "#{helm} uninstall #{release_name}"
     cmd = "#{cmd} -n #{namespace}" if namespace
+    cmd = "#{cmd} --wait" if wait
+  
     ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
   end
 
--- a/src/modules/kubectl_client/modules/modules.cr
+++ b/src/modules/kubectl_client/modules/modules.cr
@@ -79,16 +79,13 @@ module KubectlClient
       ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
     end
 
-    def self.file(file_name : String, namespace : String? = nil, wait : Bool = false)
+    def self.file(file_name : String, namespace : String? = nil, wait : Bool = true)
       logger = @@logger.for("file")
       logger.info { "Delete resources from file #{file_name}" }
 
       cmd = "kubectl delete -f #{file_name}"
       cmd = "#{cmd} -n #{namespace}" if namespace
-      if wait
-        cmd = "#{cmd} --wait=true"
-        logger.info { "Waiting until requested resource is deleted" }
-      end
+      cmd = "#{cmd} --wait=#{wait}"
 
       ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
     end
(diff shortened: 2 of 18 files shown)


src/modules/helm/helm.cr
```diff
@@ -278,13 +278,15 @@ module Helm
     ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
   end
 
-  def self.uninstall(release_name, namespace = nil) : CMDResult
+  def self.uninstall(release_name : String, namespace : String? = nil, wait : Bool? = nil) : CMDResult
```
martin-mat, line 281: imo optional should be true or false as a default. Introducing a 3rd value - nil - just brings confusions what is the default.

src/modules/kubectl_client/modules/modules.cr
```diff
@@ -79,16 +79,13 @@ module KubectlClient
       ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }
     end
 
-    def self.file(file_name : String, namespace : String? = nil, wait : Bool = false)
+    def self.file(file_name : String, namespace : String? = nil, wait : Bool? = nil)
```
martin-mat, line 82: dtto.
Just note that "wait" is by default true for kubectl delete.

src/modules/helm/helm.cr
```diff
@@ -278,13 +278,15 @@ module Helm
...
     logger = Log.for("uninstall")
     logger.info { "Uninstalling helm chart: #{release_name}" }
 
     helm = BinarySingleton.helm
     cmd = "#{helm} uninstall #{release_name}"
     cmd = "#{cmd} -n #{namespace}" if namespace
+    cmd = "#{cmd} --wait=#{wait}" unless wait.nil?
```
martin-mat, line 288: That is not correct syntax accordingly to helm docs:
https://helm.sh/docs/helm/helm_uninstall/

just "--wait" is correct.

src/modules/kubectl_client/modules/modules.cr
```diff
@@ -79,16 +79,13 @@ module KubectlClient
...
       logger = @@logger.for("file")
       logger.info { "Delete resources from file #{file_name}" }
 
       cmd = "kubectl delete -f #{file_name}"
       cmd = "#{cmd} -n #{namespace}" if namespace
-      if wait
-        cmd = "#{cmd} --wait=true"
-        logger.info { "Waiting until requested resource is deleted" }
-      end
+      cmd = "#{cmd} --wait=#{wait}" unless wait.nil?
```
martin-mat, line 88: simply true or false. Don't allow nil.

src/tasks/utils/cnf_installation/install_common.cr
```diff
@@ -107,12 +119,63 @@ module CNFInstall
...
+        Manifest.combine_ymls_as_manifest_string(
+          Manifest.manifest_path_to_ymls(manifest_path)
+        )
+      
+      timeout = parsed_args[:timeout]
+      all_uninstallations_successfull &&= wait_for_deployment_uninstallation(deployment_name, manifest, timeout)
+    end
+  
+    if all_uninstallations_successfull
+      stdout_success "All CNF deployments were uninstalled, some time might be needed for all resources to be down."
```
martin-mat, line 170: "down" does not sound exact, rather "removed"?

martin-mat (changes requested):

src/modules/helm/helm.cr
svteb: will default to false, which is the default behavior for helm.

martin-mat (approved):
lgtm

rich-l (approved):
lgtm

LuciaSirova (approved):
lgtm

Issue #2300, opened 2025-06-18
[BUG] Tests utilizing `workload_resource_test` may be reporting false results


svteb:
**Describe the bug**
After #2294 it was discovered that services were somehow being checked in the `liveness` and `readiness` tasks. This prompted a change/refactor to `workload_resource_test` as services are not workload resources. This issue was fixed in #2301. Throughout testing of this change a new (bigger) issue was discovered and that was the fact that many of the tests were using `workload_resource_test` incorrectly and have been reporting faulty results _for years_.

Many tasks in testsuite rely on the `workload_resource_test` to evaluate whether a test is successful. The usual pattern is like this:

```crystal
task "some_task"
  task_response = CNFManager.workload_resource_test(args, config) do |resource, container, initialized|
    result = some_test_for_resource(resource)
    result # issue arises if this is not at the end of the block
  end

  if task_response == true:
    good_result
  else
    bad_result
  end
end
```

The issue is that the `task_response` is rubbish in most cases as the `workload_resource_test` gets evaluated as true if a bool is not the last returned value in the block, the present code uses this pattern for block evaluation:

```crystal
def self.workload_resource_test(...)
  test_passed = true

  iterate_over_resources:
    resp = yield ...  # <= resp will store the result of block evaluation (bool?)  
    test_passed = false if resp == false # test_passed will remain true if resp == nil
  end

  initialized && test_passed # will be true even if nil was passed for all resources
end
```

An example task that does not return a bool at the end of `workload_resource_test` block is `task "pod_network_latency"` (or any other test in `src/tasks/workload/reliability.cr`). `Pod_network_latency` will always evaluate as true, not taking into account the actual result of the test.

**To Reproduce**
1. Probes issue
- deploy any service and run `liveness` or `readiness` task
- The test will tell you that there are no containers defined for service thus no probes can exist.
2. The bigger issue
- Could be checked by finding some counter to `workload/reliability.cr` tests (which we coincidentally don't have).

**Resolution**
The `workload_resource_test` will get its long-overdue refactor, removing services from there altogether, while a new function will grant the support for those tasks that need to iterate over services (or any other resources kinds for that matter). The bigger issue with the tests reporting rubbish will be resolved in a second PR as it is currently unknown what are all the affected tests and whether the spec tests will even be passing.

PR #2305, opened 2025-06-26
Fix: `workload_resource_test` PART 2/2

svteb:
## Description
- Enforces proper usage of `workload_resource_test` (block iteration now requires Bool instead of Bool?)
- Prior implementation allowed for some tasks (mostly in reliability) to be completely disfunctional (always reporting the task result as passed). More info in the referenced issue.
- All the other tasks that utilized `workload_resource_test` and weren't using it in the intended way were subtly refactored.

**Some tests now utilize early exists, that is why git diff displays indentation changes as full on code changes**

Throughout the refactoring I noticed that certain tasks could do with a further touch-up, also some tasks that utilize `workload_resource_test` don't necessarily have a binary result that could be covered with a Bool, some have skips/special cases. In the future we could consider implementing some auxiliary result-reporting logic, i.e. having `workload_resource_test` return some enum like this instead of bool:

```crystal
enum TestOutcome
  Pass
  Fail(message : String)
  Skip(reason  : String)
end

task "some_task"
  task_result = workload_resource_test(...) do |...|
    if (some_prerequisite_condition_unfulfilled)
      next TestOutcome::Skip.new("No pods for #{kind}/#{name}") # currently covered through some external variable
    
    if (some_check_true)
      next TestOutcome::Pass # currently true

    TestOutcome::Fail.new("Container does not use specialized init system...") # currently false
  end
end

final = aggregate_outcomes(task_result)
CNFManager::TestCaseResult.new(final.status, final.msg)
```

Although that would introduce a split in tests (some basing task result on bool, others on a more granular structure). Well, this is just something to consider in the future.

## Issues:
Refs: #2300

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [ ] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/utils/cnf_manager/cnf_manager.cr
+++ b/src/tasks/utils/cnf_manager/cnf_manager.cr
@@ -60,6 +60,8 @@ module CNFManager
   end
 
   def self.resource_refs(args, config, resource_kinds, &block : NamedTuple(kind: String, name: String, namespace: String) -> )
+    logger = Log.for("resource_refs")
+    logger.info { "Yielding resources: #{resource_kinds}" }
     kinds_filter = resource_kinds.map(&.downcase)
 
     cnf_resources(args, config) do |resource|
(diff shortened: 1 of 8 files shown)


collivier (approved):

martin-mat (approved):
lgtm

rich-l (approved):
lgtm

PR #2346, opened 2025-11-28
Optimize the gate duration

collivier:
It spits ["cnf_installation"], ["observability"] and ["shared_database"]
to optimize the gates. All the jobs end before 13 min now.

It also reduces the retries when searching image digests to
avoid wasting too much time in the negative testing.

It renames shared_database2 to shared_database_flaky to avoid breaking
the numbering.


rich-l (approved):
lgtm

martin-mat (approved):
great job!

PR #2351, opened 2025-12-08
Uninstall all after all the spec tests

### LINKED ISSUE #2350: [BUG] Fix failures in gates when deleting the clusters
They are always 2 to 5 jobs which exist with issues when deleting the clusters
See https://github.com/lfn-cnti/testsuite/actions/runs/19939513624/job/57196486975

it maybe be related to cluster-tools whichs aren't deleted by the spec tests as they should
cnf-testsuite        pod/cluster-tools-rz7wd                                                          1/1     Running   0              44s
cnf-testsuite        pod/cluster-tools-thndx                                                          1/1     Running   0              44s

collivier:
It may avoid some crashes when deleting kind clusters.

Fixes #2350


Smitholi67 (approved):
lgtm

martin-mat (approved):
lgtm

PR #2352, opened 2025-12-12
Call kind delete cluster until success delete

collivier:
Docker sometimes fails deleting workers:
  - Container failed to exit within 10s of kill - trying direct SIGKILL
  - cannot remove container "kind-worker": could not kill container: tried to kill container, but did not receive an exit event

kind delete cluster is an idempotent operation, meaning it may be called multiple times without failing (like "rm -f").
If the cluster resources exist they will be deleted, and if the cluster is already gone it will just return success.


Smitholi67 (approved):
lgtm

PR #2355, opened 2025-12-15
Debug remaining containers after cnf_installation2

collivier:
It adds logs to debug the next issues which will help protecting vs them in a 2nd PR.
According to /var/log/syslog, it should be the same issues as seen when deleting the kind clusters

`Container failed to exit within 10s of kill - trying direct SIGKILL`


Smitholi67 (approved):
lgtm

PR #2359, opened 2026-02-09
Update crinja to 0.9.0

collivier:
crinja 0.8.1 now fails:
```
In lib/crinja/src/runtime/resolver.cr:58:32

 58 | return Value.new object[name.to_s]
                             ^
Error: no overload matches 'Proc(Crinja::Arguments, Crinja::Value)#[]' with type String

Overloads are:
 - Proc(*T, R)#[](*args : *T)
```

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -6,7 +6,7 @@ shards:
 
   crinja:
     git: https://github.com/straight-shoota/crinja.git
-    version: 0.8.1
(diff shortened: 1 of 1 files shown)


Smitholi67 (approved):
lgtm

martin-mat (approved):
lgtm

PR #2373, opened 2026-04-08
Free5gc sigterm_handled test optimization

RobinPopov:
sigterm_handled test optimization

## Description
Free5gc sigterm_handled test optimization achieved by:

- Bug fixing the pods_by_labels() function. It now correctly matches pods based on all labels, not just by last label.
- Clearing pods_by_resource_labels() TODO. It now checks selector labels instead of general labels, resulting in narrower pod selection
- Optimization in sigterm_handled test. Skips already tested pods, if previous precautions fail. Also avoids testing multiple instances of scaled pods

## Issues:
Refs: https://github.com/lfn-cnti/testsuite/issues/2345

The issue was not in not detecting MongoDB pod. The test would get to it eventually in about 17hours.
Also the test was not looping, just checking virtually every single pod for all pods, resulting in exponential runtime growth with every added pod. The runtime was now cut to about 1 hour 20 mins from previous 18 hours, by checking every pod only once. Runtime might be machine dependent.

## How has this been tested:
 - [*] Covered by existing integration testing
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [*] Bug fix (non-breaking change which fixes an issue)
 - [ ] Documentation update

## Checklist:
**Documentation**
- [*] No updates required.

**Code Review**

**Issue**

### DIFF
--- a/src/tasks/workload/microservice.cr
+++ b/src/tasks/workload/microservice.cr
@@ -556,6 +556,9 @@ task "sig_term_handled" do |t, args|
       test_reason: String | Nil
     )
 
+    #Track already tested pods
+    tested_pods = Set(String).new
+ 
     # Iterate over all resources
     task_response = CNFManager.workload_resource_test(args, config, check_containers: false) do |resource, container, _|
       kind = resource["kind"].downcase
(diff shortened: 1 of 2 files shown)


collivier (approved):

src/tasks/workload/microservice.cr
```diff
@@ -583,6 +586,15 @@ task "sig_term_handled" do |t, args|
...
         pod_name      = pod.dig("metadata", "name").as_s
         pod_namespace = pod.dig("metadata", "namespace").as_s
+
+        # Skip already tested pods
+        pod_unique_id = "#{pod_namespace}/#{pod_name}"
+        if tested_pods.includes?(pod_unique_id)
+          logger.info { "Skipping already tested pod: #{pod_unique_id}" }
+          next true
+        end
+        tested_pods << pod_unique_id
```
martin-mat: I think the pod should be added to tested_pods only after it is really tested (it can be skipped when not up)

martin-mat (changes requested):
please check my comment

martin-mat (approved):
lgtm

PR #2375, opened 2026-04-16
Cover CNTi testsuite on arm64

collivier:
It verifies CNTi testsuite vs both amd64 clusters and arm64 clusters.

Please note that CNTi testsuite currently fails vs arm64 clusters which forces returning true.

### DIFF
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -1,0 +1,107 @@
+name: arm64
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - '**'
+  pull_request:
+jobs:
+  arm64:
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - run: |
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-arm64
+          chmod +x ./kind
+          sudo mv ./kind /usr/local/bin/kind
+          cat << EOF > /tmp/cluster.yml
+          kind: Cluster
+          apiVersion: kind.x-k8s.io/v1alpha4
+          nodes:
+            - role: control-plane
+              extraMounts:
+                - containerPath: /var/lib/kubelet/config.json
+                  hostPath: $HOME/.docker/config.json
+            - role: worker
+              extraMounts:
+                - containerPath: /var/lib/kubelet/config.json
+                  hostPath: $HOME/.docker/config.json
+            - role: worker
+              extraMounts:
+                - containerPath: /var/lib/kubelet/config.json
+                  hostPath: $HOME/.docker/config.json
+          EOF
+          kind create cluster --name arm64 --config=/tmp/cluster.yml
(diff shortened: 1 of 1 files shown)


.github/workflows/arm64.yml
```diff
@@ -0,0 +1,107 @@
...
+            - role: worker
+              extraMounts:
+                - containerPath: /var/lib/kubelet/config.json
+                  hostPath: $HOME/.docker/config.json
+          EOF
+          kind create cluster --name qemu-arm64 --config=/tmp/cluster.yml
+          rm /tmp/cluster.yml
+      - uses: actions/checkout@v4
+      - run: |
+            sudo apt update && sudo DEBIAN_FRONTEND=noninteractive apt install qemu-user-static -y
```
martin-mat, line 90: It looks that this allows amd64 binaries to be executed inside the container as well. That is probable reason why the tests pass. If it is really so, we need to make the environment more strict. Can you please double check?

.github/workflows/arm64.yml
```diff
@@ -0,0 +1,107 @@
...
+          sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash
+      - run: |
+          shards install
+          crystal build src/cnf-testsuite.cr
+          ./cnf-testsuite setup
+          ./cnf-testsuite cnf_install cnf-config=example-cnfs/coredns || true
+          ./cnf-testsuite cert
+          ./cnf-testsuite uninstall_all
+        continue-on-error: true
```
martin-mat, line 54: It would be good to fail in case that testsuite crashes on "ERROR".

.github/workflows/arm64.yml
collivier: I don't think so. But yes I can double check if it helps

.github/workflows/arm64.yml
collivier: I don't get the point as it is permissive on purpose to let the gate run.
The next patch fixing testing against arm64 clusters should remove continue on error and || true.

.github/workflows/arm64.yml
martin-mat: I can confirm my suspection:
```
# ls /.dockerenv
/.dockerenv
# uname -m
aarch64
# file /bin/bash
/bin/bash: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=bea6a154d9a9158114ee0a2a439045596615df14, for GNU/Linux 3.7.0, stripped
# file ./kubescape-ubuntu-latest
./kubescape-ubuntu-latest: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=-Kl36bDLGl5cyLKjJLpq/Tcnqkz6EiOslnmWftV4a/xS04mFsr8y0I0f-haKa9/P6XGzqWzUUaEAsI_3Pds, stripped
# ./kubescape-ubuntu-latest version
 ⚠️   current version 'v3.0.8' is not updated to the latest release: 'v3.0.15'
Your current version is: v3.0.8
```

collivier: Yes but this test still makes sens to cover arm64 -> amd64 even if it doesn't help find the current bug.

martin-mat (approved):
lgtm

PR #2376, opened 2026-04-16
Cover CNTi testsuite vs all Debian releases

collivier:
It covers CNTi Test Suite against all active Debian releases from bullseye to sid (unstable).

Please note they are 3 cases:
- no crystal nor shards package available
- only crystal package is available
- both packages are available (stable and newer)

https://packages.debian.org/fr/trixie/crystal
https://packages.debian.org/fr/trixie/shards

It runs on ubuntu-latest but the same logics would work vs v1.0.0 if any quota is reached.
https://docs.github.com/en/actions/reference/limits

### DIFF
--- a/.github/workflows/debian.yml
+++ b/.github/workflows/debian.yml
@@ -1,0 +1,80 @@
+name: debian
+on:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - '**'
+  pull_request:
+jobs:
+  debian:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        release:
+          - bullseye
+          - bookworm
+          - trixie
+          - forky
+          - sid
+    steps:
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/debian.yml
```diff
@@ -0,0 +1,80 @@
...
+              rm -rf crystal-1.19.1-1 crystal-1.19.1-1-linux-x86_64.tar.gz ;; \
+            *) \
+                DEBIAN_FRONTEND=noninteractive apt install crystal shards -y ;; \
+            esac && \
+            curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && \
+            install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl && \
+            curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-4 | bash && \
+            git config --global --add safe.directory /testsuite
+            (cd testsuite && shards install && crystal build src/cnf-testsuite.cr && \
+            ./cnf-testsuite setup && ./cnf-testsuite cnf_install cnf-config=example-cnfs/coredns && ./cnf-testsuite cert && ./cnf-testsuite uninstall_all)"
```
martin-mat, line 76: I think we need a way how to identify whether testsuite execution was OK or not. We can start with detection whether an ERROR occurred during "cnf-testsuite cert".

martin-mat: Well it will fail I guess when ERROR occurs, it should return 2.

martin-mat (approved):
lgtm

PR #2392, opened 2026-07-03
Fix upload urls for assets

collivier:
it must be uploads.github.com/repos/lfn-cnti/testsuite


collivier:
I must merge it to see if it fixes our release issues.

PR #2400, opened 2026-07-10
Fix grammar and spelling mistakes in markdown docs

collivier:


martin-mat (approved):
Incredible how many mistakes were there. Thanks.

PR #2403, opened 2026-07-23
fix: handle missing Helm during setup

### LINKED ISSUE #2383: Enhance helm missing output when calling cnf-testsuite setup
**Describe the bug**

./cnf-setup shouldn't print the missing dependencies in a full crystal backtrace.
See kubectl missing as a good output

```
./cnf-testsuite setup
CNF TestSuite version: main-2026-05-12-121232-080bef98
Successfully created directories for cnf-testsuite
src/modules/helm/utils.cr:90:18 in 'get'
src/tasks/setup/helmenv_setup.cr:10:9 in '->'
lib/sam/src/sam/task.cr:54:39 in 'call'
lib/sam/src/sam/execution.cr:20:7 in 'invoke'
lib/sam/src/sam/task.cr:44:29 in 'call'
lib/sam/src/sam/execution.cr:20:7 in 'invoke'
lib/sam/src/sam.cr:35:5 in 'invoke'
lib/sam/src/sam.cr:53:7 in 'process_tasks'
src/cnf-testsuite.cr:122:3 in '__crystal_main'
/usr/lib/crystal/lib/crystal/main.cr:118:5 in 'main_user_code'
/usr/lib/crystal/lib/crystal/main.cr:104:7 in 'main'
/usr/lib/crystal/lib/crystal/main.cr:130:3 in 'main'
/lib/x86_64-linux-gnu/libc.so.6 in '??'
/lib/x86_64-linux-gnu/libc.so.6 in '__libc_start_main'
./cnf-testsuite in '_start'
???
No Helm binary found locally or globally.
```

```
./cnf-testsuite setup
CNF TestSuite version: main-2026-05-12-121232-080bef98
Successfully created directories for cnf-testsuite
No Global kubectl version found
No Local kubectl version found
Kubectl not found

    Linux installation instructions for Kubectl can be found here: https://kubernetes.io/docs/tasks/tools/install-kubectl/

    Install kubectl binary with curl on Linux
    Download the latest release with the command:

    curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
    To download a specific version, replace the $(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt) portion of the command with the specific version.

      For example, to download version v1.18.0 on Linux, type:

      curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.18.0/bin/linux/amd64/kubectl
    Make the kubectl binary executable.

      chmod +x ./kubectl
    Move the binary in to your PATH.

      sudo mv ./kubectl /usr/local/bin/kubectl
    Test to ensure the version you installed is up-to-date:

      kubectl version --client
    
Global helm found. Version: 4.1.4
Local helm not found
Global git found. Version: 2.47.3
No Local git version found
Dependency installation failed. Some prerequisites are missing. Please install all of the prerequisites before continuing.
```

w3lld1:
## Description

Running `setup` without Helm currently calls `Helm::Binary.get` before the local installer can run, which exposes a Crystal backtrace instead of installing the missing dependency. I changed the setup task to use the existing non-raising installation check and added a regression test that runs with Helm intentionally absent from `PATH`.

## Issues

Fixes #2383

## How has this been tested

- [x] Built `src/cnf-testsuite.cr` with Crystal 1.14.0
- [x] Passed the focused missing-Helm regression spec
- [x] Passed the existing local Helm installation spec
- [x] Passed Crystal formatter and `git diff --check`
- [ ] Full Kubernetes-backed test matrix (deferred to CI)

## Types of changes

## Checklist

**Documentation**

- [x] No updates required.

**Code Review**

- [x] The missing dependency path exits through the existing installer result handling rather than an unhandled exception.

### DIFF
--- a/spec/modules/helm_spec.cr
+++ b/spec/modules/helm_spec.cr
@@ -10,6 +10,29 @@ module Helm
 end
 
 describe "Helm" do
+  describe "missing" do
+    it "installs Helm instead of exposing a missing-binary backtrace", tags: ["helm"] do
+      path_without_helm = File.tempname("path-without-helm")
+      FileUtils.mkdir_p(path_without_helm)
+      tar = Process.find_executable("tar") || raise "spec requires tar in PATH"
+      gzip = Process.find_executable("gzip") || raise "spec requires gzip in PATH"
+      File.symlink(tar, File.join(path_without_helm, "tar"))
+      # GNU tar resolves gzip through PATH when extracting the Helm archive.
+      File.symlink(gzip, File.join(path_without_helm, "gzip"))
+      Helm.uninstall_local_helm
+
+      result = ShellCmd.run_testsuite("setup:install_local_helm", "PATH=#{path_without_helm}")
+
+      result[:status].success?.should be_true
+      File.exists?(Setup::HELM_BINARY).should be_true
+      result[:output].should_not contain("No Helm binary found")
+      Dir.glob(File.join(Setup::HELM_DIR, "helm-*.tar.gz")).should be_empty
+    ensure
+      FileUtils.rm_rf(path_without_helm.not_nil!)
+      Helm.uninstall_local_helm
+    end
+  end
+
   describe "global" do
     before_all do
       Helm.uninstall_local_helm
--- a/src/modules/helm/utils.cr
+++ b/src/modules/helm/utils.cr
@@ -51,6 +51,13 @@ module Helm
 
   # Public helpers
 
+  def self.binary_found? : Bool
+    result = ToolChecker::Result.new(tool_name)
+    global_check(result)
+    local_check(result)
+    result.global_ok || result.local_ok
+  end
+
   # Temporary solution for task prereqs/specs
   def self.installation_found? : Bool
     result = check
@@ -78,8 +85,8 @@ module Helm
   class Binary
     @@helm : String = ""
 
-    # This will return path to helm binary if its found globally or installed by testsuite.
-    # If helm is not found it will return empty string.
+    # Returns the path to a global or testsuite-managed Helm binary.
+    # Raises HelmBinaryNotFoundError when neither installation exists.
     def self.get : String
       return @@helm unless @@helm.empty?
 
@@ -100,31 +107,36 @@ module Helm
     end
   end
 
-  def self.install_local_helm : Bool
+  def self.install_local_helm(force : Bool = false) : Bool
     logger = Log.for("install_local_helm")
     logger.info { "Installing Helm tool locally" }
 
     FileUtils.mkdir_p(Setup::HELM_DIR)
 
-    if File.exists?(Setup::HELM_BINARY)
+    if File.exists?(Setup::HELM_BINARY) && !force
       logger.notice { "Helm binary found in: #{Setup::HELM_BINARY}, skipping installation" }
       return true
     end
 
+    helm_archive = File.join(Setup::HELM_DIR, "helm-#{Setup::HELM_VERSION}.tar.gz")
     begin
-      helm_archive = File.join(Setup::HELM_DIR, "helm-#{Setup::HELM_VERSION}.tar.gz")
       download_file(Setup::HELM_URL, helm_archive)
     rescue ex : Exception
       logger.error { "Error while downloading Helm binary: #{ex.message}" }
       return false
     end
 
-    unless (res = TarClient.untar(helm_archive, Setup::HELM_DIR))[:status].success?
+    res = begin
+      TarClient.untar(helm_archive, Setup::HELM_DIR)
+    ensure
+      File.delete?(helm_archive)
+    end
+
+    unless res[:status].success?
       logger.error { "Error while extracting Helm binary: #{res[:error]}" }
       return false
     end
 
-    File.delete(helm_archive)
     true
   end
 
--- a/src/tasks/setup/helmenv_setup.cr
+++ b/src/tasks/setup/helmenv_setup.cr
@@ -7,12 +7,12 @@ namespace "setup" do
     logger = SLOG.for("install_local_helm")
     logger.info { "Installing Helm tool" }
 
-    if !Helm::Binary.get.empty? && !ENV.has_key?("force_install")
+    if !ENV.has_key?("force_install") && Helm.binary_found?
       logger.notice { "Helm installation has been found on the system, skipping" }
       next
     end
 
-    unless Helm.install_local_helm
+    unless Helm.install_local_helm(force: ENV.has_key?("force_install"))
       stdout_failure("Task 'install_local_helm' failed")
       exit(1)
     end


collivier:
@w3lld1 workloads fail because of missing docker secrets in your fork. We must create a local branch here instead.

w3lld1:
Understood. I don't have permission to create a branch in `lfn-cnti/testsuite`; please create one from this PR's head, and I'll continue there if any follow-up is needed.

src/tasks/setup/helmenv_setup.cr
```diff
@@ -7,7 +7,7 @@ namespace "setup" do
     logger = SLOG.for("install_local_helm")
     logger.info { "Installing Helm tool" }
 
-    if !Helm::Binary.get.empty? && !ENV.has_key?("force_install")
+    if Helm.installation_found? && !ENV.has_key?("force_install")
```
martin-mat: `Helm.installation_found?` isn't a drop-in for the old silent check — it's the *prereqs printer*: it emits `stdout_success`/`stdout_warning` lines and its `post_checks` shells out to `helm list` against the live cluster. Consequences:

- A plain `./cnf-testsuite setup` now prints the helm status block **twice** (here, then again from `setup:prereqs`) and makes two `helm version` + `helm list` round-trips.
- Every task that depends on `setup:install_local_helm` (`cnf_install`, `platform`, `install_opa`, `helm_chart_published`, `helm_chart_valid`) now prints "Global helm found… / Local helm not found" (yellow warning) before doing anything — on a global-only helm install users will read that recurring warning as an error.
- The old `Binary.get` primed the `@@helm` memo as a side effect; the new guard discards the check result, so the first later `Binary.get` repeats the whole probe including the cluster call.

The non-printing form already exists: `Helm.check.ok?` (used by this PR's own spec context). Even better, key it on binary presence only — `Helm.check` result's `local_ok || global_ok` — and leave the printing to `setup:prereqs`, whose job that is (see review body for why `ok?` specifically is a hazard here).

martin-mat: Two `force_install` notes while this guard is being touched:

1. Crystal evaluates left-to-right, so `force_install=1` still pays for the full probe (including the `helm list` cluster call) and prints "Global helm found" right before installing anyway. Swapping the operands — `!ENV.has_key?("force_install") && Helm.installation_found?` — short-circuits all of it for free and reads truer to intent.
2. Pre-existing but worth knowing: even when the guard is bypassed, `Helm.install_local_helm` early-returns `true` when `Setup::HELM_BINARY` exists (`utils.cr:109-112`) — so `force_install` can never actually repair a corrupt local helm; it prints "Helm tool has been installed" without doing anything. If force is meant to work, the early return needs to honor it too.

spec/modules/helm_spec.cr
```diff
@@ -10,6 +10,26 @@ module Helm
...
+      FileUtils.mkdir_p(path_without_helm)
+      File.symlink(Process.find_executable("tar").not_nil!, File.join(path_without_helm, "tar"))
+      File.symlink(Process.find_executable("gzip").not_nil!, File.join(path_without_helm, "gzip"))
+      Helm.uninstall_local_helm
+      Helm::Binary.clear_cache
+
+      result = ShellCmd.run_testsuite("setup:install_local_helm", "PATH=#{path_without_helm}")
+
+      result[:status].success?.should be_true
+      result[:output].should_not contain("HelmBinaryNotFoundError")
```
martin-mat: This assertion is very likely vacuous: `HelmBinaryNotFoundError` is a *class name* that doesn't appear in the exception message (`"No Helm binary found locally or globally."`, `utils.cr:96-99`), and the top-level handler prints the message plus backtrace frames — where the class name only shows up if the backtrace is symbolized (it isn't for `--release --static` binaries). So this line can pass even against the pre-fix code, and only the `success?` check above actually guards the regression.

Also, exit 0 alone doesn't prove an install happened — the skip branch and `install_local_helm`'s early-return branch also exit 0. Suggest making the assertions load-bearing:

```crystal
result[:status].success?.should be_true
File.exists?(Setup::HELM_BINARY).should be_true
result[:output].should_not contain("No Helm binary found")
```

spec/modules/helm_spec.cr
```diff
@@ -10,6 +10,26 @@ module Helm
 end
 
 describe "Helm" do
+  describe "missing" do
+    it "installs Helm instead of exposing a missing-binary backtrace", tags: ["helm"] do
+      path_without_helm = "spec/fixtures/path-without-helm-#{Process.pid}"
```
martin-mat: Two artifact/hygiene issues with this scratch setup:

1. **The test leaves a ~17 MB tarball untracked in the repo root.** The real install path runs `download_file(Setup::HELM_URL, ...)` which writes `helm-3.19.0.tar.gz` into the cwd and never deletes it; this `ensure` doesn't either, and `.gitignore` (`*.tar`, `cnf-testsuite*.tar*`) doesn't match it. Every spec run dirties `git status`, and on self-hosted runners these accumulate. It also means each CI run of the `helm` shard depends on get.helm.sh availability. The sonobuoy installer (`sonobuoy_setup.cr:33-43`) already models the fix: download into the tool dir, delete after extract.
2. **The scratch dir lives inside the tracked `spec/fixtures/`.** A hard kill (SIGKILL/CI cancellation) before the `ensure` leaves `path-without-helm-<pid>` with dangling symlinks in a tracked directory, and PID reuse makes `File.symlink` raise `File::AlreadyExistsError`. `File.tempname`/`Dir.tempdir` gives you an absolute, out-of-repo path — which also fixes the relative-PATH issue below.

spec/modules/helm_spec.cr
```diff
@@ -10,6 +10,26 @@ module Helm
...
+      FileUtils.mkdir_p(path_without_helm)
+      File.symlink(Process.find_executable("tar").not_nil!, File.join(path_without_helm, "tar"))
+      File.symlink(Process.find_executable("gzip").not_nil!, File.join(path_without_helm, "gzip"))
+      Helm.uninstall_local_helm
+      Helm::Binary.clear_cache
+
+      result = ShellCmd.run_testsuite("setup:install_local_helm", "PATH=#{path_without_helm}")
```
martin-mat, line 24: `PATH=spec/fixtures/path-without-helm-<pid>` is a *relative* PATH entry, resolved against each exec'ing process's own cwd. It's load-bearing in a subtle way: `TarClient.untar` runs `tar -xvf ... -C <target>`, and GNU tar auto-decompresses the `.gz` by exec'ing `gzip` resolved through this PATH — that's the only reason the `gzip` symlink exists at all (worth a comment, or the next reader deletes it). If tar spawns the decompressor after its `-C` chdir (implementation-dependent), or specs are ever run from a different working directory, `gzip` becomes unresolvable and the test fails with a misleading "task failed". `File.expand_path` (or an absolute tempdir per the comment above) removes the whole hazard.

Smaller points: `Process.find_executable("gzip").not_nil!` dies with an opaque `NilAssertionError` on minimal images where gzip/tar aren't standalone binaries — a guarded `raise "spec requires tar and gzip in PATH"` would fail informatively. And the two `Helm::Binary.clear_cache` calls in this example are no-ops — the install happens in the `./cnf-testsuite` subprocess, and this spec process never calls `Binary.get` (the other describes already reset their own state in `before_all`).

martin-mat (commented):
Thanks — the core fix is real: `Helm::Binary.get` raises `HelmBinaryNotFoundError` rather than returning `""` (its doc comment lies), so the old `!Helm::Binary.get.empty?` guard could never take the false branch safely and produced exactly the #2383 backtrace. Swapping it for a predicate is the right idea. Inline comments cover the details on the replacement predicate and the spec; two things can't be anchored inline because they live outside the diff:

**1. The #2383 crash survives on five other entry points.** `HelmBinaryNotFoundError` is rescued nowhere in the codebase, and these tasks reach `Helm::Binary.get` without any `setup:install_local_helm` dependency: `cnf_uninstall` (via `helm_deployment_manager.cr:45`), `setup:install_jaeger`, `setup:install_fluentd`/`install_fluentbit`, `install_kind` (`kind_setup.cr:101`), and `platform:resilience`. On a helm-less machine each of them still dumps the same backtrace this PR fixes for `setup`. A single rescue at a choke point (the generic `rescue` in `CNFManager::Task.task_runner`, or the entrypoint handler in `src/cnf-testsuite.cr`) converting it to a "helm not found — run `cnf-testsuite setup`" message + exit 1 would close the issue for all callers at once. Also worth fixing the doc comment on `Binary.get` (`utils.cr:76-95`), which still claims it returns an empty string — that's how this bug pattern gets written.

**2. The new guard's semantics rest on a latent bug.** `installation_found?` returns `Result#ok?` = `errors.empty? && (global_ok || local_ok)` — narrower than the old binary-presence check. Today they coincide only because `helm_gives_k8s_warning?` (`utils.cr:137-151`) returns *both* of its messages in the warning slot of its `{warning, error}` tuple, so `errors` is always empty (and `installation_found?`'s `stdout_failure` branch is dead code). The "Helm was found, but it could not be executed successfully" string reads like an error, so someone will eventually "fix" the tuple order — and from that day, a machine with a working global helm but no reachable cluster (post_checks runs `helm list`, and this task runs before any cluster exists in `setup`'s order) flips to `installation_found? == false` and re-downloads a local helm on every dependent task run, while `setup:prereqs` exits 1 on the same condition. Keying the guard on binary presence (`local_ok || global_ok`) instead of `ok?` decouples it.

Overall: direction is good; the main asks are a silent predicate in the guard, a load-bearing assertion in the spec, and spec artifact cleanup. The choke-point rescue could also be a follow-up issue if you prefer to keep this PR minimal.

w3lld1:
Thanks — addressed the review in `92adcf28`:

- the setup guard now uses a silent binary-only predicate, avoids the cluster post-check, and short-circuits entirely for `force_install`;
- forced installation now replaces an existing local Helm binary instead of returning early;
- the archive is downloaded under the Helm tools directory and removed after extraction;
- the regression spec uses an absolute temporary PATH, checks the installed binary and user-facing error text, and verifies archive cleanup;
- the stale `Binary.get` return-value comment now documents the exception.

I kept the broader choke-point rescue out of this focused PR, as suggested for a follow-up.

Validation:
- `crystal build --no-codegen --warnings none spec/modules/helm_spec.cr` (Crystal 1.6.2)
- `crystal tool format --check src/tasks/setup/helmenv_setup.cr spec/modules/helm_spec.cr` (Crystal 1.6.2)
- `git diff HEAD^ --check`

I did not run the runtime spec locally because its helper first builds the complete `cnf-testsuite` binary; the focused type-check above completed successfully.

src/modules/helm/utils.cr
```diff
@@ -100,26 +107,32 @@ module Helm
...
     rescue ex : Exception
       logger.error { "Error while downloading Helm binary: #{ex.message}" }
       return false
     end
 
-    unless (res = TarClient.untar(helm_archive, Setup::HELM_DIR))[:status].success?
+    res = begin
+      TarClient.untar(helm_archive, Setup::HELM_DIR)
+    ensure
+      File.delete(helm_archive) if File.exists?(helm_archive)
```
martin-mat: simplify: File.delete?(helm_archive)

w3lld1: Simplified in `7594919c` by using `File.delete?(helm_archive)` directly.

Validation: `git diff HEAD^ --check`. The focused Crystal spec could not start in this checkout because shard dependencies are not installed, so I did not treat it as passing.

martin-mat (approved):
lgtm, thanks.

martin-mat:
@w3lld1 after discussion in CNTi's TSC: Unfortunatelly we are not currently able to run CI workflows from cloned repos. Please confirm that you are interested and I will add you as a member to the organization. You will then be able to push the change to repo's branch and create a new PR.

collivier:
> @w3lld1 after discussion in CNTi's TSC: Unfortunatelly we are not currently able to run CI workflows from cloned repos. Please confirm that you are interested and I will add you as a member to the organization. You will then be able to push the change to repo's branch and create a new PR.

@w3lld1 please rebase this PR on top of main. The patch https://github.com/lfn-cnti/testsuite/commit/2e89b3f9c88d04771a6f2ea5396cc111271fcce3 was needed to allow verifying forked-PR

collivier:
> > @w3lld1 after discussion in CNTi's TSC: Unfortunatelly we are not currently able to run CI workflows from cloned repos. Please confirm that you are interested and I will add you as a member to the organization. You will then be able to push the change to repo's branch and create a new PR.
> 
> @w3lld1 please rebase this PR on top of main. The patch [2e89b3f](https://github.com/lfn-cnti/testsuite/commit/2e89b3f9c88d04771a6f2ea5396cc111271fcce3) was needed to allow verifying forked-PR

I did

martin-mat:
Blocker on the current head (`3620291`) — and to be clear up front, this isn't from your work: it came in with the rebase.

`src/modules/helm/utils.cr:140` deletes an archive the new `ensure` block already deleted eight lines earlier:

```crystal
res = begin
  TarClient.untar(helm_archive, Setup::HELM_DIR)
ensure
  File.delete?(helm_archive)     # ← deleted here
end

unless res[:status].success?
  logger.error { "Error while extracting Helm binary: #{res[:error]}" }
  return false
end

File.delete(helm_archive)        # ← line 140: raises, the file is gone
true
```

`File.delete?` returns `false` on a missing file; `File.delete` raises `File::NotFoundError`. So every successful local Helm install now ends in an unhandled backtrace — the same failure mode this PR removes from the missing-binary path, reappearing on the install path.

This is what fails `Crystal Specs (helm)`:

```
Unhandled exception: Error deleting file:
  '/root/.cnf-testsuite/tools/helm/helm-3.19.0.tar.gz' (File::NotFoundError)
  from src/modules/helm/utils.cr:140:5 in 'install_local_helm'
  from spec/modules/helm_spec.cr:69:7 in 'run_before_all_hooks'
```

It fires twice: once in the subprocess, failing the new regression test's `success?` assertion, and once in-process from the pre-existing `describe "local"` `before_all`, which aborts the file. The shard reports `5 examples, 1 failures` followed by `Aborted!`, so `local`'s four existing examples never run at all.

Where it came from — the file at each SHA:

| commit | line 140 | |
|---|---|---|
| `92adcf2` (reviewed 08-12) | `true` | ✅ |
| `7594919` (the `File.delete?` simplification) | `true` | ✅ |
| `3620291` (rebase, 08-21) | `File.delete(helm_archive)` | 🔴 |

The rebase re-applied main's trailing `File.delete` on top of the branch's new `ensure`, keeping both sides. Nothing else in the diff moved. The earlier CI runs give no cover here — the pre-rebase `helm` shards died at `##[error]Username and password required`, the forked-PR secrets problem that 2e89b3f fixed, so they never reached the specs.

**Fix: drop line 140.** The branch is also one commit behind main, so a rebase is needed anyway — that's the natural moment.

My 08-12 approval predates all of this and shouldn't be read as covering this head.

w3lld1:
Thanks — the rebased head `b6280221` removes the duplicate archive deletion while retaining cleanup in the `ensure` block. The complete reported CI matrix, including `Crystal Specs (helm)`, is green on this head.

collivier (approved):

martin-mat (approved):

Issue #2394, opened 2026-07-06
[BUG] Harden release_manager.cr


collivier:
**Describe the bug**

upload_release_asset failed silently when the upload url was fully incorrect and returned true https://github.com/lfn-cnti/testsuite/commit/9679ddf3eb3dbec01b0c93c820c8afa10002d665

- all errors must be caught and treated as such
- curl should be replaced by Crystal http client library

panagiotisevaggelou:
Hi @collivier, I have submitted a fix for this issue in PR #2404.

The fix replaces the previous `curl`-based release asset upload with Crystal's `HTTP::Client`, uses the `upload_url` returned by the GitHub release API, handles upload failures explicitly, and improves error logging.

I also verified the change locally with a mock GitHub upload server since the existing `release_manager_spec` tests are currently disabled and need some cleanup first.

martin-mat:
@collivier can we close this one?

martin-mat:
Closing as completed by #2404 (merged 2026-08-23) plus a follow-up:

- **curl → Crystal HTTP client**: `upload_release_asset` now uses `HTTP::Client.post` with the API-provided `upload_url`, streaming the file with an explicit `Content-Length`.
- **all errors caught and treated as such**: non-2xx responses raise with status and (truncated) body; `already_exists` (422) on a re-run is absorbed deliberately so `upsert_release` stays idempotent; a missing `upload_url` in the release payload raises with the API's own error instead of a bare `KeyError`.
- The silent-success mode described here (bad upload URL returning true) is exactly what the `response.success?` check now prevents.

One residual hardening landed separately: `MIME.from_filename` raising on extensionless assets was fixed with the two-arg default form. Remaining nice-to-have (re-enabling `release_manager_spec.cr`) is tracked by the TODO in that file; the fetch/upsert paths still use Halite but no longer misreport failure.

PR #2404, opened 2026-07-28
Fix GitHub release asset upload URL handling

panagiotisevaggelou:
## Description
This PR fixes the release asset upload implementation by replacing the previous `curl`-based upload with a native Crystal HTTP request.

The previous implementation relied on:

```bash
curl --http1.1 ... "https://uploads.github.com/repos/.../releases/#{release_id}/assets?name=$(basename #{asset_name})"
```

The GitHub release API returns an `upload_url` field in the release object with a URI template format:

```bash
https://uploads.github.com/repos/{owner}/{repo}/releases/{release_id}/assets{?name,label}
```

The implementation now receives the `upload_url` directly from the release object, removes the URI template part (`{?name,label}`), and appends the asset name query parameter before performing the upload.

In addition to switching to `HTTP::Client.post`, this PR also makes a small improvement to the upload flow:

- the upload logic now logs failures with additional context before re-raising the exception. (please let me know if this is considered redundant since HTTP upload failures are already handled with explicit error checks).

### Example

Before:

```bash
https://uploads.github.com/repos/.../assets{?name,label}
```

After processing:

```bash
https://uploads.github.com/repos/.../assets?name=sample.bin
```

The upload now uses:

- `HTTP::Client.post`
- `Authorization: Bearer <GITHUB_TOKEN>`
- `Content-Type: application/gzip`
- binary file contents as the request body

### Testing

I did not add an official spec test yet because the existing release_manager_spec.cr tests are currently disabled.

The existing TODO explains that the release manager tests need fixing first.

I tested the implementation locally using a small mock GitHub upload server instead of the real GitHub API.

The mock server simulated the GitHub release asset endpoint and logged:

- received URL and query parameters
- authorization header
- content type
- content length
- uploaded binary contents
- returned a successful GitHub-like JSON response

Test setup:

```bash
GITHUB_TOKEN=dummy-token crystal run src/modules/release_manager/test_upload_asset.cr
```

Mock server:

```bash
python3 mock_github_upload.py
```

Local test output:

```bash
[2026-07-27 17:03:55] starting upload test
[2026-07-27 17:03:55] asset_name: sample.bin
[2026-07-27 17:03:55] original upload_url: http://127.0.0.1:4567/repos/lfn-cnti/testsuite/releases/123/assets{?name,label}
[2026-07-27 17:03:55] asset exists?: true
[2026-07-27 17:03:55] asset size: 20 bytes
[2026-07-27 17:03:55] final upload url: http://127.0.0.1:4567/repos/lfn-cnti/testsuite/releases/123/assets?name=sample.bin
[2026-07-27 17:03:55] GITHUB_TOKEN present?: yes
[2026-07-27 17:03:55] calling ReleaseManager.upload_release_asset(...)
[2026-07-27 17:03:55] upload returned successfully
{
 "id" => 123,
 "name" => "sample.bin",
 "browser_download_url" => "http://127.0.0.1:4567/download/sample.bin"
}
```

```bash
[2026-07-27 17:03:55] === POST received ===
[2026-07-27 17:03:55] path:   /repos/lfn-cnti/testsuite/releases/123/assets
[2026-07-27 17:03:55] query:  {'name': ['sample.bin']}
[2026-07-27 17:03:55] auth:   Bearer dummy-token
[2026-07-27 17:03:55] type:   application/gzip
[2026-07-27 17:03:55] len:    20
[2026-07-27 17:03:55] body bytes read: 20
[2026-07-27 17:03:55] body preview: b'hello release asset\n'
[2026-07-27 17:03:55] responding with 201 success JSON
```

## Issues:
Refs: #2394

## How has this been tested:
 - [ ] Verified all A/C passes
     * [ ] develop
     * [ ] master
     * [ ] tag/other branch
 - [ ] Test environment
    * [ ] New Packet K8s cluster
    * [ ] Kind cluster
 - [ ] Have not tested

## Types of changes:
 - [ ] Documentation update

## Checklist:
**Documentation**
- [x] No updates required.

**Code Review**

**Issue**


collivier:
@panagiotisevaggelou Thank you for your contribution. Could you please rebase your proposal on top of https://github.com/lfn-cnti/testsuite/pull/2402 (if it's still makes sense) which was just merged.

collivier:
@panagiotisevaggelou workloads fail because of missing docker secrets in your fork. We must create a local branch here instead.

panagiotisevaggelou:
Hi @collivier, thanks for pointing this out. I see that #2402 already implemented the main part of the fix. I was not aware of that change when I started working on the issue. I rebased my branch on top of the latest changes so the remaining differences between the two implementations can be compared more easily. If the remaining changes are not needed, feel free to close this PR.

src/modules/release_manager/release_manager.cr
```diff
@@ -210,27 +210,29 @@ module ReleaseManager
...
+        upload_url,
+        headers: headers,
         body: File.read(asset_name)
       )
-      client.close
-      asset_resp = response.body
-      Log.info {"asset_resp: #{asset_resp}"}
-      asset = JSON.parse(asset_resp.strip)
+      unless response.success?
+        raise "Asset upload failed: #{response.status_code} #{response.body}"
```
martin-mat: **Blocker: this breaks re-runs of the release workflow.** GitHub's upload endpoint returns **422 `already_exists`** when an asset with the same name is already attached to the release. Previously that error body was parsed and returned, so re-running `upsert_release` for a tag whose asset already uploaded (e.g. retrying CI after a later step failed) succeeded. Now the 422 raises, propagates out of the SAM task, and the Release job fails on **every** retry until someone manually deletes the asset — the exact case an *upsert* exists to absorb. A transient failure also now leaves a published (`draft: false`) release with notes but no binary, and the retry can't self-heal.

The raise itself is an improvement over silently reporting success — but please special-case the duplicate-asset path, e.g. on 422 check the body for `already_exists` and return/log instead of raising.

Minor on the same line: interpolating the full `response.body` into the exception message can embed an entire HTML error page; consider truncating (`response.body[0, 512]`).

martin-mat (commented):
Thanks for the fix — the direction is right on both counts: using the API-provided `upload_url` removes the hardcoded `lfn-cnti/testsuite` upload host (the old code ignored the configured repo), and checking `response.success?` fixes failed uploads being silently reported as success.

That said, the rewrite introduces one significant regression (the `already_exists` re-run case, see inline comment on the `raise`) and drops two safeguards the old code had (MIME detection and `File.basename`/URL-encoding of the asset name) — details inline.

One more thing that can't be anchored inline: the commented-out example call sites at lines ~156–160 still show the old signature (`upload_release_asset(found_release["id"], path, name)`), which is now incompatible with the new method. Anyone resurrecting that block would pass a release id where an upload URL is expected — it compiles fine and only fails at runtime during a release. Suggest deleting the dead block (git history preserves it) or updating the signatures.

src/modules/release_manager/release_manager.cr
```diff
@@ -210,27 +210,29 @@ module ReleaseManager
...
-          "Authorization" => "Bearer #{ENV["GITHUB_TOKEN"]}",
-          "Content-Type" => mime_type,
-        },
+  def self.upload_release_asset(upload_url, asset_name)
+    # TODO Add test that checks for uploaded corrupted binary.
+    begin
+      upload_url = "#{upload_url.split("{").first}?name=#{asset_name}"
+      headers = HTTP::Headers{
+        "Authorization" => "Bearer #{ENV["GITHUB_TOKEN"]}",
+        "Content-Type" => "application/gzip",
```
martin-mat: Hardcoding `application/gzip` regresses the previous `MIME.from_filename(asset_name) || "application/octet-stream"`. It's wrong today for the `CRYSTAL_ENV=TEST` path: the tarball step is skipped there and the raw `cnf-testsuite` ELF binary is uploaded — now recorded in the GitHub asset metadata as `application/gzip`. Consumers selecting assets by `content_type`, or browsers/proxies trusting the header, get a mislabeled artifact. Restoring the one-line MIME derivation keeps the helper correct for all assets.

src/modules/release_manager/release_manager.cr
```diff
@@ -210,27 +210,29 @@ module ReleaseManager
...
-      response = client.post(
-        "#{uri.path}?name=#{asset_name_base}",
-        headers: HTTP::Headers{
-          "Authorization" => "Bearer #{ENV["GITHUB_TOKEN"]}",
-          "Content-Type" => mime_type,
-        },
+  def self.upload_release_asset(upload_url, asset_name)
+    # TODO Add test that checks for uploaded corrupted binary.
+    begin
+      upload_url = "#{upload_url.split("{").first}?name=#{asset_name}"
```
martin-mat: Two safeguards got dropped here:

1. **`File.basename` is gone** — `asset_name` doubles as the local file path (`File.read` below), so any caller passing a path sends slashes inside `?name=`, which GitHub rejects with 422 (now a hard failure). The commented-out call sites above (~156–160) pass exactly such paths.
2. **No URL-encoding** — the version string derives from the git branch name, so a branch containing `#` (legal in git) truncates the query as a URI fragment and the asset silently uploads under a wrong name with no `.tar.gz` extension; `&` injects a bogus query param.

Suggest: `upload_url = "#{upload_url.split("{").first}?#{URI::Params.encode({"name" => File.basename(asset_name)})}"`

Also consider assigning to a new local instead of reassigning the `upload_url` parameter — after this line the original argument is gone, which makes the error log below misleading when debugging.

src/modules/release_manager/release_manager.cr
```diff
@@ -210,27 +210,29 @@ module ReleaseManager
...
+      headers = HTTP::Headers{
+        "Authorization" => "Bearer #{ENV["GITHUB_TOKEN"]}",
+        "Content-Type" => "application/gzip",
+      }
+      response = HTTP::Client.post(
+        upload_url,
+        headers: headers,
         body: File.read(asset_name)
```
martin-mat: Pre-existing, but since this method is being rewritten anyway: `File.read` slurps the whole tarball (tens of MB) into memory as a String. `HTTP::Client.post` accepts an `IO` body, so you can stream it — just set an explicit `Content-Length` header (GitHub's uploads endpoint rejects chunked transfer encoding):

```crystal
headers["Content-Length"] = File.size(asset_name).to_s
File.open(asset_name) do |file|
  response = HTTP::Client.post(upload_url, headers: headers, body: file)
  ...
end
```

src/modules/release_manager/release_manager.cr
```diff
@@ -210,27 +210,29 @@ module ReleaseManager
...
-      asset_resp = response.body
-      Log.info {"asset_resp: #{asset_resp}"}
-      asset = JSON.parse(asset_resp.strip)
+      unless response.success?
+        raise "Asset upload failed: #{response.status_code} #{response.body}"
+      end
+      asset = JSON.parse(response.body)
       Log.info {"asset: #{asset}"}
       asset
+    rescue ex
```
martin-mat: This `rescue ex; Log.error; raise ex` only logs and rethrows — the exception already propagates with its message intact, so every failure is now reported twice (and for the raise above, the duplicated message contains the full response body). Suggest dropping the `begin/rescue` wrapper entirely, or using a method-level `rescue` if the log line is wanted.

src/modules/release_manager/release_manager.cr
```diff
@@ -162,7 +162,7 @@ module ReleaseManager
       ## const assets = [bin_asset_resp, source_tarball_asset_resp, source_zip_asset_resp]
 
       Log.info {"uploading binary"}
-      asset = ReleaseManager.upload_release_asset(found_release["id"], cnf_bin_asset_name)
+      asset = ReleaseManager.upload_release_asset(found_release["upload_url"].as_s, cnf_bin_asset_name)
```
martin-mat: Switching to `upload_url` is a real fix (the old code hardcoded `lfn-cnti/testsuite` as the upload host, ignoring `@repo_name`). But `found_release["upload_url"].as_s` reads a field from a response that's never status-checked — if the preceding POST/PATCH failed (limited-permission `GITHUB_TOKEN` → 403, deleted release → 404, rate limit), the error body is `{"message": ...}` and this line crashes with a bare `KeyError: Missing hash key: "upload_url"`, hiding the real API error. The old `["id"]` had the same hazard, so this is carried over rather than new — but a cheap guard here (`found_release["upload_url"]?` plus a descriptive raise) would make the failure mode obvious.

collivier:
@panagiotisevaggelou  please rebase this PR on top of main. The patch https://github.com/lfn-cnti/testsuite/commit/2e89b3f9c88d04771a6f2ea5396cc111271fcce3 was needed to allow verifying forked-PR

collivier:
> @panagiotisevaggelou please rebase this PR on top of main. The patch [2e89b3f](https://github.com/lfn-cnti/testsuite/commit/2e89b3f9c88d04771a6f2ea5396cc111271fcce3) was needed to allow verifying forked-PR

I did

panagiotisevaggelou:
> > @panagiotisevaggelou please rebase this PR on top of main. The patch [2e89b3f](https://github.com/lfn-cnti/testsuite/commit/2e89b3f9c88d04771a6f2ea5396cc111271fcce3) was needed to allow verifying forked-PR
> 
> I did

Hi @collivier, @martin-mat,

I am currently inspecting all the review comments and working on applying the requested fixes. I will push an updated commit shortly!

panagiotisevaggelou:
Hi @collivier, @martin-mat,

I have updated the branch and addressed all the review feedback:

- Handled the "asset already exists" error gracefully so re-running a failed workflow won't crash if the file is already uploaded.

- Restored file name cleaning with File.basename, proper URL encoding with URI::Params.encode, and automatic file type detection with MIME.from_filename.

- Used File.open to upload the file directly from disk instead of loading the whole file into memory first.

- Added a safety check for the upload URL with clearer error messages, shortened long error outputs in logs, cleaned up unnecessary error-handling boilerplate, and removed old commented-out code.

- Added require "mime" and require "uri" at the top so every used standard library is explicitly declared.

Testing & Commits:

- Tested all paths locally using a mock server (normal upload, duplicate asset retry, and server error handling).

- Ran crystal tool format and placed formatting updates into a separate commit to keep the main logic easy to review.

```
evaggelo@PANAGIOTIS-PC:~/cnti/testsuite$ python3 src/modules/release_manager/mock_github_upload.py
[2026-08-21T12:53:55] starting mock server on http://127.0.0.1:4567
[2026-08-21T12:54:07] === POST received ===
[2026-08-21T12:54:07] path:        /repos/lfn-cnti/testsuite/releases/123/assets
[2026-08-21T12:54:07] query:       {'name': ['sample.bin']}
[2026-08-21T12:54:07] auth:        Bearer mock-token-12345
[2026-08-21T12:54:07] content-type:application/octet-stream
[2026-08-21T12:54:07] len header:  20
[2026-08-21T12:54:07] body read:   20 bytes
[2026-08-21T12:54:07] body sample: b'hello release asset\n'
[2026-08-21T12:54:07] -> responding with 201 success
[2026-08-21T12:54:07] === POST received ===
[2026-08-21T12:54:07] path:        /repos/lfn-cnti/testsuite/releases/123/assets
[2026-08-21T12:54:07] query:       {'name': ['duplicate.bin']}
[2026-08-21T12:54:07] auth:        Bearer mock-token-12345
[2026-08-21T12:54:07] content-type:application/octet-stream
[2026-08-21T12:54:07] len header:  24
[2026-08-21T12:54:07] body read:   24 bytes
[2026-08-21T12:54:07] body sample: b'duplicate release asset\n'
[2026-08-21T12:54:07] -> responding with 422 already_exists
[2026-08-21T12:54:07] === POST received ===
[2026-08-21T12:54:07] path:        /repos/lfn-cnti/testsuite/releases/123/assets/fail
[2026-08-21T12:54:07] query:       {'name': ['sample.bin']}
[2026-08-21T12:54:07] auth:        Bearer mock-token-12345
[2026-08-21T12:54:07] content-type:application/octet-stream
[2026-08-21T12:54:07] len header:  20
[2026-08-21T12:54:07] body read:   20 bytes
[2026-08-21T12:54:07] body sample: b'hello release asset\n'
[2026-08-21T12:54:07] -> responding with 500 error
```

```
evaggelo@PANAGIOTIS-PC:~/cnti/testsuite$ crystal run src/modules/release_manager/test_upload_asset.cr
current_branch during compile: "fix/harden-release-asset-upload"
current_tag during compile: 
[2026-08-21 12:54:07] --- TEST 1: Standard Upload ---
[2026-08-21 12:54:07] Result 1:
{"id" => 123,
 "name" => "sample.bin",
 "browser_download_url" => "http://127.0.0.1:4567/download/sample.bin"}
[2026-08-21 12:54:07] --- TEST 2: Duplicate Upload (422 already_exists) ---
[2026-08-21 12:54:07] Result 2 (should return JSON without throwing):
{"message" => "Validation Failed",
 "errors" => [{"resource" => "ReleaseAsset", "code" => "already_exists"}]}
[2026-08-21 12:54:07] --- TEST 3: Simulated 500 Failure ---
[2026-08-21 12:54:07] Success: Caught expected error -> Asset upload failed: 500 - {"message":"Internal Server Error Simulation"}
```

collivier:
@panagiotisevaggelou thank you. we would need to rebase it on top of https://github.com/lfn-cnti/testsuite/pull/2449

panagiotisevaggelou:
> @panagiotisevaggelou thank you. we would need to rebase it on top of #2449

Done!

collivier (approved):

src/modules/release_manager/release_manager.cr
```diff
@@ -200,37 +178,45 @@ module ReleaseManager
...
-      asset = JSON.parse(asset_resp.strip)
-      Log.info {"asset: #{asset}"}
-      asset
-  end
+  def self.upload_release_asset(upload_url : String, asset_name : String)
+    # TODO Add test that checks for uploaded corrupted binary.
+    base_name = File.basename(asset_name)
+    query_params = URI::Params.encode({"name" => base_name})
+    formatted_upload_url = "#{upload_url.split('{').first}?#{query_params}"
+    mime_type = MIME.from_filename(asset_name) || "application/octet-stream"
```
martin-mat: Correction to my own earlier review — I steered this one wrong, so flagging it now.

Back on 08-11 I asked to restore `MIME.from_filename(asset_name) || "application/octet-stream"`, calling it "correct for all assets." It isn't: on a filename **with no extension**, `MIME.from_filename` **raises `KeyError`** rather than returning `nil`, so the `|| "application/octet-stream"` fallback never runs. Verified on both Crystal 1.6.2 (the release toolchain, `release.yml` pins `crystal:1.6.2-alpine`) and 1.10:

```crystal
MIME.from_filename("cnf-testsuite")                          # => raises KeyError
MIME.from_filename("cnf-testsuite") || "application/octet-stream"  # => still raises
MIME.from_filename("cnf-testsuite", "application/octet-stream")    # => "application/octet-stream"
```

The no-extension asset is reachable: under `unless ENV["CRYSTAL_ENV"]? == "TEST"` the tarball step is skipped, so `cnf_bin_asset_name` stays the bare `cnf-testsuite` ELF and this line raises before the upload happens.

**Severity: latent, not a live release-breaker.** The real release path (`v*` tags, no `CRYSTAL_ENV=TEST`) builds `cnf-testsuite-<version>.tar.gz` → `.gz` → `application/gzip`, which is fine. It only bites the TEST path, currently reachable via the disabled `release_manager_spec.cr`. But it's a one-word fix, and the code explicitly branches on `CRYSTAL_ENV=TEST` so that path is meant to work:

```crystal
mime_type = MIME.from_filename(asset_name, "application/octet-stream")
```

This is a pre-existing bug the rewrite carries over, not something you introduced — but since the line is being rewritten anyway (and my earlier note pointed you at the broken form), the two-arg overload closes it for good.

martin-mat:
One branch-hygiene note before merge — the PR currently carries a **stray `node_drain` commit** that isn't part of this fix.

`bb0410fe "Stop node_drain cordoning one node while draining another"` sits under the two release commits on this branch. That work landed on `main` separately via #2449 (as `e33a34f1`), but because it was squash/rebase-merged to a **different SHA**, git doesn't see `bb0410fe` as already-merged. So the PR's diff against `main` currently includes changes it shouldn't:

```
src/tasks/workload/state.cr        | 212 +++++++++++++------------
src/tasks/utils/litmus_manager.cr  |  37 ++++-
docs/TEST_DOCUMENTATION.md         |   2 +-
```

Worse, `bb0410fe` is a *stale* version of that work — it differs from what actually merged to `main` by ~85 lines (the final #2449 head had post-review fixes). Merging as-is would partially revert `node_drain`.

A clean rebase onto current `main` should fix it — the `node_drain` commit should drop out since its content is already upstream. Because of the squash-merge SHA mismatch, git may surface a conflict on `state.cr`/`litmus_manager.cr` instead of dropping it silently; if so, resolve in favour of `main` (`git checkout origin/main -- src/tasks/workload/state.cr src/tasks/utils/litmus_manager.cr docs/TEST_DOCUMENTATION.md`). After the rebase the PR should show only `release_manager.cr` (plus the formatting commit).

Everything else here is solid — the `upload_url` switch, the streaming upload with explicit `Content-Length` (correctly avoids chunked, which GitHub's uploads endpoint rejects — I verified on 1.6.2), the `already_exists` handling, and the URL-encoding all check out. Just the MIME one-liner above and this rebase.

panagiotisevaggelou:
@martin-mat @collivier Thanks for the detailed review and for catching the `MIME.from_filename` fallback behavior, appreciate the explanation!

- Updated the MIME lookup to `MIME.from_filename(asset_name, "application/octet-stream")` to safely handle extensionless binaries.

- Rebased cleanly on top of the latest main to drop the merged #2449 commit. The PR now only touches release_manager.cr.

martin-mat (approved):

PR #2410, opened 2026-08-11
spec: warn when a CNF is already installed before running specs

### LINKED ISSUE #2100: [Feature] Warning for running specs with cnfs present
**Is your feature request related to a problem? Please describe.**
Spec tests behave incorrectly in some cases when any CNFs are already installed. That could cause unexpected and hardly traceable errors on the spec run.

**Describe the solution you'd like**
I propose printing a warning for the user during the initial stages of the spec run (probably somewhere before the Testsuite build). This will help avoid such situations and provide the user with information so they can react accordingly.

martin-mat:
## Description

As proposed in the issue: spec tests behave incorrectly in some cases when a CNF is already installed, causing hard-to-trace errors. This adds an early warning in `spec/spec_helper.cr` — before the testsuite build step — when the `installed_cnf_files` directory (`CNF_DIR`) is present:

```
WARN - A CNF appears to be installed ('installed_cnf_files' directory present). Spec tests assume a clean environment and may behave unexpectedly. Consider running './cnf-testsuite cnf_uninstall' first.
```

Notes:
- The check is on directory presence rather than `CNFManager.cnf_installed?` so it also catches leftover artifacts from a partially removed installation, and costs nothing at spec startup.
- Warning only, as the issue suggests — the run is not aborted, so intentional setups (e.g. specs that expect a pre-installed CNF) keep working.
- Compile-checked with `crystal build --no-codegen spec/spec_helper.cr`.

## Issues

Fixes #2100

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -13,8 +13,13 @@ require "../src/modules/k8s_netstat"
 require "../src/modules/cluster_tools"
 require "../src/modules/kubectl_client"
 
-ENV["CRYSTAL_ENV"] = "TEST" 
+ENV["CRYSTAL_ENV"] = "TEST"
 
+# Specs assume no CNF is installed; a leftover installation can make them fail
+# in ways that are hard to trace back to the environment.
+if Dir.exists?(CNF_DIR)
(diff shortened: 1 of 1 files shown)


rich-l (approved):
lgtm

PR #2418, opened 2026-08-11
fix: forward CLI args from cert tasks to sub-tests

### LINKED ISSUE #2412: [BUG] cert task drops all CLI arguments for its sub-tests
**Describe the bug**

The `cert` task silently drops all CLI arguments for the tests it runs. `invoke_tasks_by_tag_list` (`src/tasks/cert/cert_utils.cr:28-35`) calls `parent_task.invoke(task)` without forwarding `Args`, which hits `Sam::Execution#invoke(name, *args)` (`lib/sam/src/sam/execution.cr:31-33`) and constructs an **empty** `Args` for the sub-task.

As a result, under `cert`:

- `strict` is ignored (checked per-test in `single_task_runner`, `task.cr:105`),
- `cnf-config=` is ignored,
- feature-level flags `poc`/`wip`/`alpha`/`beta` and `destructive` are ignored,
- every `~exclusion` is ignored.

Only `exclude=` works, because it is handled separately by `cert_utils.cr:3-25` before invocation.

**To reproduce**

`./cnf-testsuite cert strict` — a failing essential test does not stop the run, unlike `./cnf-testsuite workload strict`.

**Proposed fix**

Forward the parent task's `Args` through `invoke_tasks_by_tag_list` to each sub-task invocation.

**Acceptance criteria**

- `cert strict` stops on first failure like other aggregates.
- `cert ~<test_name>` excludes the test.
- Spec covering at least the `strict` and `~exclusion` forwarding.

martin-mat:
## Description

`invoke_tasks_by_tag_list` (`src/tasks/cert/cert_utils.cr`) invoked each cert sub-test with `parent_task.invoke(task)`, which lands on Sam's `invoke(name, *args)` overload and constructs an **empty** `Args`. Every CLI argument was therefore silently dropped for tests running under `cert`/`cert_*`: `strict`, `cnf-config=`, `poc`/`wip`/`alpha`/`beta`, `destructive`, and all `~exclusions`. Only `exclude=` worked (handled separately before invocation).

The fix threads the parent task's `args` through `invoke_tasks_by_tag_list` to each sub-test invocation (the `cert` aggregate itself already forwards args to its declared deps since its block arity is 2, so this was the only gap). All seven `cert_*` call sites updated.

## Verification

Verified against a live binary with an unreachable `KUBECONFIG` (so nothing can touch a cluster):

- **pre-fix binary**: `cert_state ~node_drain` → exclusion dropped, `node_drain` invoked (fails on kubeconfig), exit 1
- **fixed binary**: same command → zero sub-tests run, `State results: 0 of 1 tests passed`, exit 0

Added a cluster-free regression spec (tag `points`, existing CI shard): `node_drain` is the only cert-tagged state test, so `cert_state ~node_drain` must run zero sub-tests and succeed; before the fix the dropped exclusion made `node_drain` abort the run with "You must install a CNF first".

## Issues

Fixes #2412

### DIFF
--- a/spec/utils/cnf_manager_spec.cr
+++ b/spec/utils/cnf_manager_spec.cr
@@ -85,7 +85,16 @@ describe "SampleUtils" do
   end
 
   it "'points_yml' should parse and return the points yaml file", tags: ["points"]  do
-    (CNFManager::Points.points_yml.find {|x| x["name"] =="liveness"}).should be_truthy 
+    (CNFManager::Points.points_yml.find {|x| x["name"] =="liveness"}).should be_truthy
+  end
+
+  it "cert tasks should forward CLI args to their sub-tests", tags: ["points"] do
+    # node_drain is the only cert-tagged state test, so excluding it must leave
+    # cert_state with nothing to run. Without arg forwarding the exclusion is
+    # dropped and node_drain aborts the run because no CNF is installed.
+    result = ShellCmd.run_testsuite("cert_state ~node_drain")
+    result[:status].success?.should be_true
+    result[:output].should_not contain("You must install a CNF first")
   end
 
   it  "'task_points' should return the amount of points for a passing test", tags: ["points"] do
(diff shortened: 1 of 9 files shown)


rich-l (approved):
lgtm

PR #2420, opened 2026-08-11
Make points.yml handling lazy and warn on silent CWD-dependent behavior

### LINKED ISSUE #2413: [BUG] CWD-dependent behavior: platform/cert crash without points.yml, missing config.yml silently changes behavior
**Describe the bug**

The binary behaves differently — or crashes — depending on the directory it is run from:

1. **`platform` and `cert` crash in a fresh directory.** `points.yml` is read with no existence check (`File.open("points.yml")`, `src/tasks/utils/cnf_manager/points.cr:191`), but only `setup` and `workload` depend on `setup:configuration_file_setup`, the task that writes it (`src/tasks/setup/setup.cr:47-50`). `platform` (`src/tasks/platform/platform.cr:3`) and `cert` (`src/tasks/cert/cert.cr:9`) do not, so scoring raises `File::NotFoundError` when `points.yml` isn't already in the CWD.
2. **Missing `./config.yml` silently changes behavior.** `BASE_CONFIG = "./config.yml"` (`src/tasks/constants.cr:9`) is CWD-relative; when absent, `toggle()` (`src/tasks/utils/utils.cr:84-85`) silently returns false for `wip`/`poc`/`alpha`/`beta`/`destructive`, and the configured `loglevel` (`src/tasks/utils/logging.cr:74-79`) reverts to the default — with no warning.
3. **Missing points.yml entries degrade silently.** When a task has no `points.yml` entry, scoring falls back to `default_scoring` with only a log-level warning (`points.cr:231`).

**Proposed fix**

- Add `setup:configuration_file_setup` to `platform` and `cert` dependency lists — or make `Points.points_yml` lazily self-create `points.yml` from the embedded copy (it is baked into the binary via `EmbeddedFileManager`).
- Print a visible warning when `./config.yml` is absent but toggles/loglevel would have been consulted, and when scoring falls back to `default_scoring`.

**Acceptance criteria**

- `./cnf-testsuite platform` and `./cnf-testsuite cert` work from an empty directory (given a cluster).
- Running from a directory without `config.yml` produces an explicit warning instead of silent behavior change.

martin-mat:
## Description

While implementing the issue, the investigation corrected part of its premise (see the issue comment): the reported `platform`/`cert` crash on a missing `points.yml` **cannot occur**, because `src/tasks/constants.cr` called `EmbeddedFileManager.points_yml_write_file` at require time — rewriting `points.yml` into the current directory on **every invocation**, even `help` and `version`, and hard-exiting 1 at startup from a read-only directory for any command.

This PR replaces that startup side effect with lazy self-healing in `Points.points_yml`:

- materialize `points.yml` from the embedded copy when missing,
- refresh it when its content differs from the embedded copy (so a copy from another release can't linger — this preserves the always-in-sync property the startup write provided),
- memoized per process; `setup:configuration_file_setup` still writes it explicitly.

Plus the two warnings the issue asks for:

- one-time stdout warning when `./config.yml` is absent and a feature toggle is consulted (previously all toggles silently read as disabled),
- stdout warning when a test has no `points.yml` entry and falls back to `default_scoring` (previously log-only at warn level, invisible at the default error level).

## Verification (main binary vs this branch, empty scratch dirs, unreachable KUBECONFIG)

| Scenario | main | this PR |
|---|---|---|
| `version` in empty dir | leaves `points.yml` behind | no files created |
| `version` in read-only dir | **exit 1** at startup | works, exit 0 |
| `cert_state exclude=node_drain` in empty dir | works (startup write) | works (lazy write), `points.yml` + `results/` created |
| stale/garbage `points.yml` in dir | refreshed at startup | refreshed on first scoring read |

Specs are unaffected: they invoke `setup:configuration_file_setup` explicitly (unchanged) or call `Points.points_yml` in-process (now self-healing). Compile-checked and behavior-verified with a built binary.

## Issues

Fixes #2413

### DIFF
--- a/src/tasks/constants.cr
+++ b/src/tasks/constants.cr
@@ -33,7 +33,6 @@ EmbeddedFileManager.chaos_network_loss
 EmbeddedFileManager.chaos_cpu_hog
 EmbeddedFileManager.chaos_container_kill
 EmbeddedFileManager.points_yml
-EmbeddedFileManager.points_yml_write_file
 EmbeddedFileManager.enforce_image_tag
 EmbeddedFileManager.constraint_template
 EmbeddedFileManager.disable_cni
--- a/src/tasks/utils/cnf_manager/points.cr
+++ b/src/tasks/utils/cnf_manager/points.cr
@@ -187,7 +187,18 @@ module CNFManager
       end
     end
 
+    @@points_yml_ensured = false
+
     def self.points_yml
+      # points.yml used to be rewritten into the CWD at require time on every
+      # invocation, even for help/version. Materialize (or refresh, when a copy
+      # from another release lingers) the embedded points.yml only when scoring
+      # actually needs it, so non-scoring commands stay free of side effects
+      # and work from read-only directories.
+      unless @@points_yml_ensured
+        create_points_yml unless File.exists?("points.yml") && File.read("points.yml") == POINTSFILE
+        @@points_yml_ensured = true
+      end
       points = File.open("points.yml") { |f| YAML.parse(f) }
       points.as_a
     end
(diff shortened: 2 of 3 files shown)


rich-l (approved):
lgtm

Issue #2416, opened 2026-08-11
[BUG] CLI arguments are not validated: typos silently ignored, values truncated at second '=', unmatched ~exclusions


martin-mat:
**Describe the bug**

CLI arguments are not validated anywhere, so mistakes are silently swallowed instead of reported:

1. **Unknown `key=value` args and bare flags are silently ignored** (`lib/sam/src/sam/args.cr:60-102` — no schema, no allowlist). `cnf_config=x`, `--cnf-config=x`, `stricT`, `essentials` all parse fine and do nothing. Live in-repo example: `spec/workload/microservice_spec.cr:161` passes `force=true`, which nothing in `src/` reads. The suite itself writes a misspelled arg `reslience_tests` (`src/tasks/workload/5g_validator.cr:24`) that is never read.
2. **Values truncated at the second `=`**: `option_value` uses `split("=")[1]` (`args.cr:90-100`), so `cnf-config=path=with=equals` becomes `path`. Should be `split("=", 2)`.
3. **A leading-dash token consumes the next token as its value**: `./cnf-testsuite all -v strict` yields `named["v"]="strict"` and `strict` disappears from the raw args (`args.cr:73,82-88`).
4. **Empty-string argument crashes**: `./cnf-testsuite all ""` raises `IndexError` (`str[0]` in `args.cr`).
5. **Numeric args crash with a backtrace on non-numeric input**: `timeout=abc` → bare `.to_i` → `ArgumentError` → top-level backtrace dump (`src/tasks/utils/cnf_installation/install_common.cr:50,61`; same for `baseline_count`, `5g_validator.cr:37`).
6. **`~exclusion` matching no task is silently ignored** (`lib/sam/src/sam/task.cr:38`) — the in-repo CI depends on three dead exclusions today (see companion CI-cleanup issue).

The SAM shard is a fork pinned by commit (`vulk/sam.cr`, `shard.lock`), so parser fixes can be made in the fork.

**Proposed fix**

- Central registry of accepted named keys and bare flags (global + per-task); unknown input → usage error with a distinct exit code and the accepted options listed.
- Fix `split("=", 2)`, the dash-token value rule, and the empty-arg crash in the SAM fork.
- Friendly error for non-numeric numeric args.
- Warn (or error) when a `~exclusion` matches no known task.

**Acceptance criteria**

- A typo'd key/flag/exclusion produces a visible error/warning, not silence.
- `cnf-config=a=b` passes `a=b` through intact.
[...]

PR #2421, opened 2026-08-11
ci: remove dead task exclusions and stop swallowing install failures

### LINKED ISSUE #2417: [MAINTENANCE] CI workflows carry dead/wrong-scope task exclusions and swallow install failures
The in-repo GitHub workflows carry task exclusions that match no task (silently ignored, because `~exclusion` is unvalidated — see companion argument-validation issue) and swallow install failures:

**1. Dead exclusions** in `.github/workflows/actions.yml:521,607,693`:
- `~increase_capacity` and `~decrease_capacity` — the actual task is `increase_decrease_capacity` (`src/tasks/workload/compatibility.cr:172`). These exclusions have never excluded anything, which means the binary-smoke jobs have been *running* increase/decrease capacity all along (possibly intentionally by now — decide and either fix the names or drop them).
- `~install_script_helm` — no such task exists anywhere.

**2. Wrong-scope exclusion**: `~resilience` excludes only the root workload `resilience` task; `platform:resilience` is a different task (namespace-qualified matching, `lib/sam/src/sam/task.cr:28,38`). Currently harmless because `~platform` is also passed, but worth a comment or the qualified form to prevent surprises if `~platform` is ever dropped.

**3. Swallowed failures**: `arm64.yml:50` runs `./cnf-testsuite cnf_install cnf-config=example-cnfs/coredns || true` — an install failure is ignored and the subsequent `cert` run fails confusingly (or tests nothing meaningful). `debian.yml:75` chains the same commands inside a shell string. Replace `|| true` with proper failure handling (fail the job with a clear message, or make the step conditional).

**Acceptance criteria**

- Every `~exclusion` in the workflows matches an existing task (guarded by the existing tag-drift check pattern, or by the upcoming exclusion validation).
- No `|| true` on install steps; an install failure fails the job with a readable error.

martin-mat:
## Description

Cleanup of the workflow invocations per the issue:

1. **Dead exclusions removed** from the binary smoke jobs in `actions.yml`: `~increase_capacity`, `~decrease_capacity` (the real task is `increase_decrease_capacity`) and `~install_script_helm` (no such task). SAM silently ignores exclusions that match nothing, so these never excluded anything — the smoke jobs have been running `increase_decrease_capacity` all along and passing. Removing them is behavior-neutral and stops the workflow from implying that test is skipped.
2. **arm64: `|| true` removed from `cnf_install`.** With the runner's default `bash -e`, an install failure now stops the step instead of proceeding to run `cert` against nothing. The step remains advisory (it has `continue-on-error: true`); whether to make the arm64 job blocking is a separate decision left untouched. The debian workflow already chains with `&&` and needed no change.

Note on `~resilience` (from the issue): exclusion matching is namespace-qualified, so `~resilience` covers only the workload resilience suite, while `platform:resilience` is separately covered by `~platform` in the same invocations. Recording that here rather than as comments in the workflow.

## Verification

- Cross-checked every remaining `~exclusion` in all workflows against the actual task definitions (including the dynamically-defined `rolling_update`/`rolling_downgrade`/`rolling_version_change` tasks) — all map to existing tasks.
- Both workflow files validated as parseable YAML.

Companion issue #2416 will add runtime validation so unmatched `~exclusions` warn instead of being silently ignored.

## Issues

Fixes #2417

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -518,7 +518,7 @@ jobs:
         ./cnf-testsuite setup
         wget -O cnf-testsuite.yml https://raw.githubusercontent.com/lfn-cnti/testsuite/${GITHUB_SHA}/example-cnfs/coredns/cnf-testsuite.yml
         ./cnf-testsuite cnf_install cnf-config=./cnf-testsuite.yml
-        LOG_LEVEL=debug ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~increase_capacity ~decrease_capacity ~install_script_helm ~helm_chart_valid ~helm_chart_published
+        LOG_LEVEL=debug ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~helm_chart_valid ~helm_chart_published
         LOG_LEVEL=debug ./cnf-testsuite cnf_uninstall
         LOG_LEVEL=debug ./cnf-testsuite uninstall_all
     - name: Delete Cluster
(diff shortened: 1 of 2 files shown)


rich-l (approved):
lgtm

PR #2423, opened 2026-08-11
Validate CLI arguments instead of silently ignoring mistakes

### LINKED ISSUE #2416: [BUG] CLI arguments are not validated: typos silently ignored, values truncated at second '=', unmatched ~exclusions

martin-mat:
## Description

Implements the issue: a pre-flight validator now runs in the entry point before `Sam.process_tasks`, closing the class of silently-swallowed CLI mistakes.

**Usage errors** (exit `64` = sysexits `EX_USAGE`, message lists the accepted inputs):
- unknown `key=value` argument (`bogus_key=1`)
- unknown bare flag (`stricT`, `essentials`)
- unknown dash option (`--bogus` — previously this cascaded into "Task -l was not found")
- non-numeric value for `timeout=`/`baseline_count=` (previously a raw backtrace)
- empty argument (previously `IndexError`)

**Warnings** (run continues — kept non-fatal for compatibility with existing scripts/CI):
- `~exclusion` matching no task ("will be ignored")
- a task name in an argument position, pointing at the `@` multi-task separator (previously `cnf-testsuite state security` silently ran only `state`)

**Parser fixes**: the vendored SAM shard (`lib/` is not committed; pinned fork `vulk/sam.cr`) is reopened in `src/tasks/utils/sam_args_patch.cr` to split values on the *first* `=` only (`cnf-config=a=b` previously became `a`) and to survive empty arguments. The file documents that it can be dropped if the fixes land upstream.

**Cleanup**: removed the misspelled dead argument write `reslience_tests` (written, read by nothing) in `5g_validator.cr` and the stray `force=true` (read by nothing) in `microservice_spec.cr`. Exit code 64 documented in USAGE.md.

The allowlists were built from a sweep of every `args.named`/`args.raw` read in `src/` (8 named args, 9 flags); all spec and workflow invocations were cross-checked against them.

## Verification

Built binary, all scenarios exercised:

| Input | Result |
|---|---|
| `version bogus_flag` / `version bogus_key=1` / `version \"\"` / `version --bogus` / `timeout=abc` | exit 64 + descriptive message |
| `version ~no_such_task` / `version delete_results` | warning printed, task still runs, exit 0 |
| `version strict ~resilience` (valid flag + matching exclusion) | silent, exit 0 |
| `Sam::Args.new([\"cnf-config=a=b\"])` | value preserved as `a=b` |

New spec `spec/utils/cli_args_spec.cr` (tag `points`, existing CI shard) covers the usage errors, both warnings, and the `=`-preservation fix.

## Issues

Fixes #2416

Related: #2417 removed the dead exclusions the in-repo CI itself carried; this PR adds the runtime warning that would have caught them.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -120,7 +120,7 @@ items:
 | `schema_version` | Integer version of this results-file schema; bumped on breaking changes. |
 | `status` | Overall run verdict, derived from `exit_code`: `passed` (0), `failed` (1), `error` (2). |
 | `command` | The command line that produced this file. |
-| `exit_code` | Process exit code: `0` passed, `1` failed (a required test failed), `2` error (a test raised). |
+| `exit_code` | Process exit code: `0` passed, `1` failed (a required test failed), `2` error (a test raised). Additionally, the process exits with `64` (usage error) on unknown or malformed command-line arguments, before any test runs. |
 | `summary` | Aggregate numbers for the whole run (see below). |
 | `items` | One entry per test that ran (see below). |
 
--- a/spec/utils/cli_args_spec.cr
+++ b/spec/utils/cli_args_spec.cr
@@ -1,0 +1,32 @@
+require "./../spec_helper"
+require "../../src/tasks/utils/sam_args_patch"
+
+describe "CLI argument validation" do
+  it "rejects unknown flags and named arguments with a usage error", tags: ["points"] do
+    result = ShellCmd.run_testsuite("version bogus_flag")
+    result[:status].exit_code.should eq(64)
+    result[:output].should contain("Unknown flag 'bogus_flag'")
+
+    result = ShellCmd.run_testsuite("version bogus_key=1")
+    result[:status].exit_code.should eq(64)
+    result[:output].should contain("Unknown argument 'bogus_key='")
+
+    result = ShellCmd.run_testsuite("cnf_install cnf-config=whatever timeout=abc")
+    result[:status].exit_code.should eq(64)
+    result[:output].should contain("not a number")
+  end
+
+  it "warns on unmatched exclusions and task names in argument positions", tags: ["points"] do
+    result = ShellCmd.run_testsuite("version ~no_such_task")
+    result[:status].success?.should be_true
+    result[:output].should contain("does not match any task")
+
+    result = ShellCmd.run_testsuite("version delete_results")
+    result[:status].success?.should be_true
+    result[:output].should contain("separate them with")
+  end
+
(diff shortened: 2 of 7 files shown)


rich-l (approved):
lgtm

PR #2424, opened 2026-08-11
Derive the exit code from whether the run met its objective

### LINKED ISSUE #2411: [BUG] Failing test runs exit with code 0
**Describe the bug**

A run full of failing tests exits with code 0, so any CI gate on the test suite's exit code is currently a no-op.

The final exit code is read back from the results YAML (`src/cnf-testsuite.cr:126-136`). `exit_code: 1` is only ever written when `CNFManager::Points.failed_required_tasks` is non-empty (`src/cnf-testsuite.cr:25-36,56-66`, `src/tasks/platform/platform.cr:17-21`, `src/tasks/cert/cert.cr:29-39`), and that helper counts a failed test only if its `points.yml` entry has `required: true` (`src/tasks/utils/cnf_manager/points.cr:515`) — **no entry in `embedded_files/points.yml` sets `required`**, so the path is dead. The only live ways to get a non-zero exit are an *errored* test (uncaught exception → exit 2, `task.cr:98-100`) or `strict` mode — and `strict` aborts the whole run on the first failure (`task.cr:105-117`), so there is no "run everything, exit non-zero at the end" mode, which is what CI wants.

**Expected behavior**

Documented exit-code contract: `0` = all tests passed (incl. skipped/na), `1` = at least one test failed, `2` = a test errored / critical failure.

**Proposed fix**

Derive `exit_code: 1` from `summary.failed > 0` in `Points.write_summary!` (keeping 2 for errors), rather than resurrecting the unused `required:` mechanism. Alternative considered: mark essential/cert tests `required: true` in `embedded_files/points.yml` — but that would leave non-essential failures still exiting 0.

**Acceptance criteria**

- A run with ≥1 failed test exits 1; with ≥1 errored test exits 2; all-pass/skip/na exits 0.
- Contract documented in USAGE.md (Results Output section).
- Spec asserting a failing run exits non-zero.
- Release notes call out the behavior change for anyone (accidentally) relying on exit 0.

martin-mat:
Closes #2411.

## Problem

Failing runs exited 0. Failures only mapped to exit 1 via the unused `required:` field in `points.yml`, so a run full of failed tests still looked successful to CI.

## Approach

The exit code is now derived from the recorded items, making it an exact function of the run's outcomes — it drops back to 0 when items are cleaned, and can't be clobbered by a later aggregate writing to the results file.

An **errored test always wins with exit 2**: the suite itself broke, which is never an acceptable result. Otherwise the exit code answers *"did this run meet its objective?"* — and the objective depends on the task group:

| run | objective | exit 0 | exit 1 |
|---|---|---|---|
| `cert` | meet the certification criterion | certified | not certified |
| `all` / `workload` / `platform` | nothing failed | no test failed | ≥1 test failed |

A group with a pass criterion is judged **by that criterion alone**. `cert` certifies against an essential-test threshold, so individual test failures are *inputs to the verdict*, not a separate failure mode — the threshold already accounts for them.

That keeps the flagship check trivial, which is the point:

```bash
./cnf-testsuite cert     # exit 0 ⟺ certified. No output parsing, no guards.
```

Any CI runner, `set -e`, or `&&` chain handles that directly, while 2-vs-1 still separates "the suite broke" from "the CNF isn't certified".

## Changes

- `Points.pass_threshold` carries the criterion; `cert` sets it, everything else leaves it `nil`.
- New derivation in `Points.write_summary!`.
- Removed the dead `failed_required_tasks` block in `cert`: its `update_yml` was immediately overwritten by the derivation, and its "Test Suite failed!" line would now contradict a certified run exiting 0.
- `task.cr`: dropped the redundant `update_yml` exit-code writes that pure derivation makes dead.
- USAGE.md exit-code contract updated.

## CI

The coredns `cert` runs in the portability workflows stay **bare** — coredns is expected to certify, so `./cnf-testsuite cert` exiting 0 is exactly the assertion those jobs want. As a result this PR leaves `arm64.yml` and `debian.yml` untouched.

The `all ~...` smoke runs in `actions.yml` keep `|| [ $? -eq 1 ]`: they have no criterion, and the example CNFs are expected to fail some of those tests.

## Verification

- Full binary builds.
- `crystal spec --tag points` — 20 examples, 0 failures, including a new spec covering both directions: a failed test no longer forces exit 1 once a criterion is met, and an unmet criterion exits 1 even with nothing failed.
- All three workflow files re-validated as YAML.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -518,7 +518,8 @@ jobs:
         ./cnf-testsuite setup
         wget -O cnf-testsuite.yml https://raw.githubusercontent.com/lfn-cnti/testsuite/${GITHUB_SHA}/example-cnfs/coredns/cnf-testsuite.yml
         ./cnf-testsuite cnf_install cnf-config=./cnf-testsuite.yml
-        LOG_LEVEL=debug ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~helm_chart_valid ~helm_chart_published
+        # The example CNF is expected to fail some tests (exit 1); only an errored test (exit 2+) fails the job
+        LOG_LEVEL=debug ./cnf-testsuite all ~compatibility ~resilience ~reasonable_startup_time ~reasonable_image_size ~platform ~helm_chart_valid ~helm_chart_published || [ $? -eq 1 ]
         LOG_LEVEL=debug ./cnf-testsuite cnf_uninstall
         LOG_LEVEL=debug ./cnf-testsuite uninstall_all
     - name: Delete Cluster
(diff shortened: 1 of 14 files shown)


collivier:
@martin-mat I would prefer to keep a boolean exit code to simplify managing CI workflows. Why not consider that an exit code of 1 could indicate a crash, a skipped or failed test run independently, or a suite where certain criteria are unmet?

martin-mat:
@collivier we already have exit codes 0, 1, 2 in current code. Multiple error codes are superset of "boolean" exit for bash - any non-zero fails. 

One think I'd reconsider though compared to current PR: for result 0 we consider "set criteria passed" (like cert passed the criterias):

- tool ok and test set criteria passed ->  result=0
- tool ok and set  set criteria failed-> result=1
- tool ok -> result<2
- tool error -> result=2
- user input error -> result=64

The main idea about this change - to comply with Unix norm for command-line tools.
See examples:
grep: 0 -> match, 1-> no match,. 2-> error. Same here.
pytest: 0->ok, 1->failures, 2->interrupted, 3->internal error, 4->usage error, 5->no tests to execute
So the Unix convention exists.

Note that CI can easily handle results as needed:
all tests need to pass -> just execute, any non-zero return will cause failure
testsuite needs to execute successfully (test failures allowed) -> ./cnf-testsuite cert || [ $? -eq 1 ] (or use allow_failure: exit_codes: [1] for gitlab)

So I'd suggest to keep the logic of this PR, the only thing to modiufy -> soften criteria for "0" -> test failures allowed up to set criteria (for example cert test passed, number of passed over cert treshold)

collivier:
@martin-mat lgtm as long as criteria passed is a success (exit 0)

collivier (approved):

PR #2426, opened 2026-08-11
Refactor Sonobuoy setup task

martin-mat:
## Description

Revival of #2257 by @rafal-lal, cherry-picked onto current main (the rest of that branch's refactor series already landed as rebased commits).

- Rewrites `install_sonobuoy`/`uninstall_sonobuoy` in the `setup` namespace with structured logging (`SLOG`) and explicit error handling, matching the already-merged kind/kubescape/helm setup task refactors.
- Replaces the inline URL and backtick `tar -xzf ... && chmod ... && rm` pipeline with the `Setup::SONOBUOY_*` constants, `download_file`, `TarClient.untar` (result checked), `File.chmod`, and archive deletion after extraction.
- Uninstall keeps the guard for a missing binary and skips removing the tools dir when `sonobuoy delete` fails, so deletion can be retried.

Changes on top of the original branch:

- Fixes `Setup::SONOBUOY_URL` (previously unused) to use underscores in the asset name — `sonobuoy_<ver>_linux_amd64.tar.gz`; the dashed `linux-amd64` variant returns 404. With the constant, the download now derives OS/arch from `TARGET_OS`/`TARGET_ARCH` instead of hardcoding `amd64`, so arm64 works.
- Drops the now-unused `SONOBUOY_OS` constant.
- The already-installed check tests for the binary rather than the directory, so a previously failed install doesn't permanently block reinstallation.

The `setup` aggregate and `tools_uninstall` already reference `setup:install_sonobuoy`/`setup:uninstall_sonobuoy`, so no dependency updates were needed.

## Issues:

Refs #2257

### DIFF
--- a/src/tasks/setup/sonobuoy_setup.cr
+++ b/src/tasks/setup/sonobuoy_setup.cr
@@ -1,60 +1,61 @@
 require "sam"
 require "file_utils"
-require "colorize"
-require "totem"
-require "http/client"
-require "halite" 
 require "../utils/utils.cr"
 
-def sonobuoy_details(cmd_path : String)
-  Log.trace { cmd_path }
-  stdout = IO::Memory.new
-  status = Process.run("#{cmd_path} version", shell: true, output: stdout, error: stdout)
-  Log.debug { stdout }
-end
+namespace "setup" do
+  desc "Sets up Sonobuoy in the K8s Cluster"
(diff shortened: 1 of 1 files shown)


rich-l (approved):
lgtm

collivier (approved):

PR #2428, opened 2026-08-18
Give every task group a declared success criterion

martin-mat:
## Core of the change

`cert` was the only group that knew whether it had passed. Every other group printed a score and left the reader to interpret it — nothing about pass or fail was machine-readable per group. **A group now declares its own pass criterion and reports its own verdict.**

```crystal
category_task "state", ["no_local_volume_configuration", "elastic_volumes",
                        "database_persistence", "node_drain"]

suite_task "cert", [...],
  scope: "essential",     # whose tests count
  min_passed: 15,         # how many must pass
  max_failed: CNFManager::NO_FAILURE_LIMIT   # the threshold accounts for failures
```

## Architecture

**A criterion is a set of cumulative thresholds** — every one declared must hold:

| threshold | default | meaning |
|---|---|---|
| `min_passed` | `0` | at least this many tests in scope passed |
| `min_ratio` | none | at least this share of the tests that **exist** in scope passed |
| `max_failed` | `0` | at most this many failed; `NO_FAILURE_LIMIT` for no limit |

Defaults mean *"nothing in this group failed"*, which is what every group has always demanded. Only `cert` declares otherwise.

Note `max_failed` is the one threshold that defaults to a *constraint* rather than to "unset". If it defaulted to no-limit, a group declaring nothing would have no criteria at all and pass unconditionally — the `all` bug below, generalised to every group.

`min_ratio` counts against how many tests are **declared** in scope, not how many a run reached: `max_passed` shrinks when tests are excluded, and `cert` takes an `exclude` argument, so a ratio over it could be met by running less.

**Declaring at the group, not centrally, is what makes the verdict reliable.** `category_task`/`suite_task` register the group and emit its verdict, so a group cannot exist without one. The same declaration says whether a group is a *category* or a *suite* — which is where a test's category comes from, replacing the hand-maintained list #2448 introduced.

## Two bugs this fixes

- **Three groups reported nothing.** The verdict used to be emitted from `stdout_score` keyed by the group's *display* name, so `compatibility`, `resilience` and `observability` looked up a criterion that did not exist and silently said nothing.
- **`all` always passed.** No test carries an `all` tag, so scoping its criterion to its own name counted zero tests, found zero failures, and passed unconditionally — and `all` is the outermost group of an `all` run, so **a run with failing tests exited 0**, exactly what #2424 exists to prevent. A criterion can now be scoped to every test that ran, and `all` is declared that way. A spec guards it: a criterion whose scope matches no tests fails CI.

## Both surfaces

stdout, in a stable greppable shape, naming whichever threshold is binding:

```
Security: FAILED (2 of 19 tests failed)
Certification: PASSED (17 of 19 essential tests passed, threshold 15)
```

and `summary.criteria` in the results file, one entry per group in evaluation order. Dependencies run before a parent's body, so the **last entry is the outermost group** and `exit_code` follows its verdict — keeping #2424's contract while making the objective explicit per group.

## Also

- Every group's body was the same three lines of reporting copied 17 times; the DSL does it now, so most groups are pure declarations. `suites.cr` went from 38 lines to two.
- `all` and `workload` move out of `src/cnf-testsuite.cr` into `src/tasks/suites.cr` — they are suites like `platform` and `cert`, and defining two of the four in the entry point meant they registered only when the binary ran.
- `cert`'s description no longer claims "some percentage of essential tests"; the criterion is a fixed count.

## Verification

- `crystal spec --tag points` — 39 examples, 0 failures.
- The scope guard verified by reverting `all` to its broken scope: it fails with the group, the scope, and why.
- A real run's `summary.criteria` validates against the updated JSON schema.

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -136,6 +136,8 @@ All counts/scores are scoped to the tests that actually ran.
 | `essential_passed` / `essential_max_passed` | Passed vs. maximum-passable among `essential`-tagged tests. |
 | `points` | Total points scored. |
 | `maximum_points` | Maximum points achievable by the tests that ran. |
+| `criteria` | Success-criterion verdict per task group that ran (see below). Absent when no such group ran. |
+
 ##### Test scoring
 
 A test declares everything the suite knows about it at its own definition:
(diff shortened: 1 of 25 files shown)


collivier:
@martin-mat As for #2424 I would prefer to keep a boolean exit code to simplify managing CI workflows. Why not consider that an exit code of 1 could indicate a crash, a skipped or failed test run independently, or a suite where certain criteria are unmet?

martin-mat:
One thing to fold in while this PR is touching `cert.cr`: the `cert` task's `desc` is now stale.

```crystal
desc "The CNF Test Suite program certifies a CNF based on passing some percentage of essential tests."
```

It surfaces verbatim in the new `cnf-testsuite help tasks` listing (#2443), where it sits next to one-line descriptions of the other suites:

```
TEST SUITES
  all         Run every workload and platform test
  workload    Run every workload test against the installed CNF
  platform    Run every test against the Kubernetes platform itself, rather than the CNF
  cert        The CNF Test Suite program certifies a CNF based on passing
              some percentage of essential tests.
```

"some percentage" also predates the concrete contract: it is a fixed count (`ESSENTIAL_PASSED_THRESHOLD`, currently 15 essential tests), not a percentage — and with #2424 the exit code now states the verdict directly.

Suggestion:

```crystal
desc "Run the certification tests; exits 0 when the CNF is certified"
```

Raising it here rather than in #2443 since this PR already owns the certification-verdict wording and would otherwise conflict on the same lines.

martin-mat:
Addressed in `0f392647`. `cert` now reads:

```crystal
desc "Run the certification tests; exits 0 when the CNF is certified"
```

which renders on one line alongside the other suites:

```
TEST SUITES
  all         Run every workload and platform test
  workload    Run every workload test against the installed CNF
  platform    Run every test against the Kubernetes platform itself, rather than the CNF
  cert        Run the certification tests; exits 0 when the CNF is certified
```

Checked for the same stale "some percentage" wording elsewhere in `src/`, `docs/` and `USAGE.md` — this was the only occurrence.

collivier (approved):

PR #2440, opened 2026-08-19
Skip Docker Hub login steps when secrets are unavailable

collivier:
So forked-PR workflows can run without failing on missing credentials.


martin-mat (approved):
lgtm

Issue #2444, opened 2026-08-20
[BUG] Tasks that belong to no suite: an essential test skipped by `workload`, a platform test registered at the root namespace, a dead points.yml entry


martin-mat:
Building the grouped `help tasks` listing (#2429 / #2443) required assigning every task to a group. Anything no group claims lands in an "Other" bucket, which turned out to be a useful canary: six tasks belong to no suite. Three of them are real defects, and one changes what a certification run measures.

Everything below was verified against the built binary and `embedded_files/points.yml`, not inferred from source.

---

### 1. `verify_secrets_encryption` is registered at the root namespace, not under `platform:`

In `src/tasks/platform/security.cr`, the `namespace "platform" do` block **closes at line 96**, but the task is defined at **line 183** — with two-space indentation that makes it look like it is still inside:

```
7:   namespace "platform" do
69:    task "verify_configmaps_encryption" do |t, args|
96:  end                                    <-- namespace ends here
183:   task "verify_secrets_encryption", ["setup:kubescape_scan"] do |t, args|
```

So the registered path is `verify_secrets_encryption`, while its sibling is `platform:verify_configmaps_encryption`. Confirmed against the real task list:

```
$ cnf-testsuite completion | tr ' ' '\n' | grep verify_
verify_secrets_encryption
platform:verify_configmaps_encryption
```

Consequences:

- It **still runs**: `platform:security` lists it as a dependency and SAM's lookup walks up to the parent namespace.
- It **cannot be invoked** as `platform:verify_secrets_encryption`, unlike every other platform test.
- **`~platform:verify_secrets_encryption` silently does nothing.** SAM matches exclusions on the task's path (`args.raw.find { |x| "~#{path}" == x }`), and the path is the bare name. This is exactly the class of silently-ignored exclusion #2416 set out to eliminate.
- Its points.yml tags say otherwise: `tags: ["platform", "platform:security", "dynamic"]` — identical to `verify_configmaps_encryption`.

Fix: move the task inside the namespace block.

### 2. `zombie_handled` is an essential test that `workload` never runs

```yaml
- name: zombie_handled
  tags: [microservice, dynamic, workload, cert, essential]
```

It is defined in `src/tasks/workload/microservice.cr:450` but is **not** in the `microservice` aggregate's dependency list. The two entry points select tests differently:

- `cert_microservice` selects **by points.yml tag** (`invoke_tasks_by_tag_list`) → **runs** `zombie_handled`
[...]

PR #2447, opened 2026-08-20
Declare a test's type and scoring at the test itself

martin-mat:
Stacked on #2428. Closes the three defects in #2444.

## The problem

`points.yml` described the same tests the task definitions did, and the two descriptions could disagree. They had:

- **`zombie_handled`** — tagged `microservice/cert/essential` but absent from the `microservice` aggregate. `cert` selects by tag, `workload` runs by dependency, so **cert ran it and workload did not** — an *essential* test, counting toward the certification threshold.
- **`smf_upf_heartbeat`** — tagged `5g` while the `5g` list omitted it, so that suite had never run it.
- **`ip_addresses`** — scored but had no task at all, inflating `max_points`/`max_passed` for a test that could never run.
- **`verify_secrets_encryption`** — defined *after* its `namespace "platform" do` block closed, so it registered at the root: never invocable as `platform:verify_secrets_encryption`, and `~platform:verify_secrets_encryption` silently excluded nothing.

## The change

A test declares only what neither the task graph nor the file layout can tell us:

```crystal
scored_task "liveness",
  type: CNFManager::TestType::Essential,
  emoji: "🥑" do |t, args|
  ...
end
```

**Type is policy** — it decides the point value and what `cert` certifies — so it has to be stated. Everything else is read back:

| fact | where it comes from |
|---|---|
| **scope** | the namespace the test registers in — never declared |
| **category** | the aggregate that runs it — dependency lists stay exactly where they were |
| **tags** | derived from the two above plus type |
| **points** | derived from type |

Membership is still stated once, in SAM's own structure, and `TestRegistry` maps a test back to its category from the task graph. The tag vocabulary the rest of the suite reads — cert selection, score totals, group criterion scope — is derived rather than written down anywhere.

## Two platform tests move into the namespace

`k8s_conformance` and `clusterapi_enabled` were the only platform tests outside the `platform:` namespace — not by design, but because `platform.cr` (which also defines the top-level `platform` suite) never had a namespace block, and the tests beside it inherited that. They were the sole reason `scored_task` needed a `scope:` escape hatch.

They are now `platform:k8s_conformance` and `platform:clusterapi_enabled`, and **both keep their original names as aliases**, the way `uninstall` aliases `cnf_uninstall`. So `scope:` is gone entirely — scope is derived in every case.

An alias needed more than a task, though. SAM matches an exclusion against a task's *own path*, so `~k8s_conformance` would have matched the alias and **silently excluded nothing** — worse than before the rename, when it at least warned that it matched no task. `alias_task` records the mapping and the entry point rewrites `~alias` to name the real task, so both spellings work. The platform specs deliberately keep invoking the historical names, which exercises the aliases on every run.

Suggestions also match on the last path segment now, since a namespaced task is nine edit-distance from its own bare name — so anything else that moves into a namespace points at its new name rather than failing blankly.

## A guard, instead of a redundant parameter

A spec pins the invariant that broke twice: **every declared test must be run by something, and no test may be claimed by two categories** (which would make its derived category ambiguous). That catches `zombie_handled` and `smf_upf_heartbeat` without any per-test category declaration.

Two documented exceptions, recorded by name so they are decisions rather than silent gaps:

- `clusterapi_enabled` — poc-gated, meaningful only when invoked by name.
- `k8s_conformance` — no category at all, being a direct member of the `platform` suite. That is fine: something runs it.

## points.yml is gone

Not just superseded — it was a genuine wart. It was written into the working directory and then **read straight back**:

```crystal
create_points_yml unless File.exists?("points.yml") && File.read("points.yml") == POINTSFILE
points = File.open("points.yml") { |f| YAML.parse(f) }
```

A provable no-op round-trip that nonetheless **required a writable working directory to read data the binary was compiled with** — and that looked like configuration while silently reverting any edit, because the write triggered precisely when the on-disk copy differed.

## What stays central

**Group success criteria**, in `Points::GROUP_CRITERIA`. Unlike a test's own metadata these are group-level policy — what the suite certifies, what it demands of each category — and belong somewhere reviewable in one piece.

## Scoring

Point value follows from the type (essential 100, normal 5, bonus 1), so no test restates its own score. The 25 tests that had **no type at all** are declared `normal` — what `default_scoring` was already giving them — as is **`smf_upf_core_validator`**, a real test that turned out to have no points.yml entry whatsoever. Every essential test was also a cert test and vice versa (identical 19-test sets), so cert membership is derived rather than declared twice.

Dead weight removed: `pass5`/`pass3` (no test can produce `ResultStatus::Pass5`/`Pass3`; the enum values and the now-purposeless `to_basic` go too), `neutral` (no reader), `required` (never set by any entry), `empji` — a typo that had left `oci_compliant` with no emoji.

## Behaviour changes (deliberate)

- **`zombie_handled`** joins `microservice` — `cert` was already running it.
- **`smf_upf_heartbeat`** joins `5g`.

## Verification

**Membership identical** across all thirteen category aggregates, checked against the built binary:

| aggregate | members | | aggregate | members |
|---|---|---|---|---|
| compatibility | 10 ✅ | | 5g | 3 ✅ |
| state | 4 ✅ | | ran | 1 ✅ |
| security | 19 ✅ | | platform:security | 5 ✅ |
| configuration | 11 ✅ | | platform:observability | 4 ✅ |
| observability | 5 ✅ | | platform:resilience | 1 ✅ |
| microservice | 8 ✅ | | platform:hardware_and_scheduling | 1 ✅ |
| resilience | 10 ✅ | | | |

- **Scoring identical**: a platform run still reports `0 of 65 points (13 tests)`.
- **No file written**: ran from a clean directory, no `points.yml` appears.
- `crystal spec --tag points` — **34 examples, 0 failures**.
- Also fixed a pre-existing display bug this surfaced: `help <task>` looked dependencies up by full path, so a namespaced aggregate showed no members. It now resolves them the way SAM does.


martin-mat:
Superseded by #2448 — same branch, same commit, different base.

Reordering the stack to #2424 → points.yml-removal → group-criteria meant this PR's base (`feat/cert-verdict`) ended up *containing* its commit, so GitHub auto-closed it as merged. Nothing was merged to `main`; the work continues in #2448 with base `feat/exit-code-on-failure`.

The reorder was worth it: the group-criteria PR (#2428) no longer has to declare its criteria table in `points.yml` and then have this PR move it to Crystal — it now lands on a codebase where `points.yml` is already gone.

PR #2454, opened 2026-08-24
Print the results-file path once per run

martin-mat:
Since #2452 the entrypoint ends every run with one stable, greppable line:

```
Results: results/cnf-testsuite-results-<ts>.yml (latest: results/latest.yml)
```

**Eight** older `Results have been saved to …` prints survived alongside it — one in the group task body ([group_task.cr](src/tasks/utils/group_task.cr)), one in each of the seven `cert_*` category tasks — so a `cert` run named the same file up to nine times. This drops them all. The entrypoint line prints on the same runs the removed lines did (an invoked task that produced a results file), and nothing greps the removed text — no spec, workflow, or doc.

## The cert_* dedup that fell out

The seven `cert_*` tasks were carrying more than the extra line: each was a copy of the same twelve-line body, differing only in category tag, section heading, and score title. Editing seven copies to remove one line was the moment to stop having seven copies: `run_cert_category` in [cert_utils.cr](src/tasks/cert/cert_utils.cr) now holds the body (heading, `exclude=` handling, the `essential` flag, tag-selected invocation, section score), and each task declares its three strings:

```crystal
task "cert_state" do |t, args|
  run_cert_category(t, args, "state", "State Tests")
end
```

Output is byte-identical apart from the dropped line — headings and score titles are passed through verbatim, including the legacy spelling differences (`Upgradability` heading vs `Upgradeability` score). Two bits of debris went with it: the empty `_cert_compatibility_title` task (referenced nowhere) and two stray trace logs in `cert_resilience`.

Net: **+25 / −102.**

## Verification

Binary and specs compiled with **Crystal 1.6.2** (the CI/release compiler):

- `crystal spec --tag points` — 49/49 (includes the `cert_state ~node_drain` arg-forwarding example, which exercises the rewritten path end to end)
- `./cnf-testsuite cert_state ~node_drain` → exactly **one** line mentioning the results file
- `./cnf-testsuite _divide_by_zero` → zero `saved to` lines, one `Results:` line

### DIFF
--- a/src/tasks/cert/cert_compatibility.cr
+++ b/src/tasks/cert/cert_compatibility.cr
@@ -8,23 +8,6 @@ require "./cert_utils.cr"
 
 desc "The CNF test suite checks to see if CNFs support horizontal scaling (across multiple machines) and vertical scaling (between sizes of machines) by using the native K8s kubectl"
 task "cert_compatibility" do |t, args|
-  puts "Compatibility, Installability & Upgradability Tests".colorize(Colorize::ColorRGB.new(0, 255, 255))
-
-  exclude = get_excluded_tasks(args)
-  essential_only = args.raw.includes? "essential"
-  tags = ["compatibility", "cert"]
-  tags << "essential" if essential_only
-
-  invoke_tasks_by_tag_list(t, args, tags, exclude_tasks: exclude)
-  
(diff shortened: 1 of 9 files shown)


rich-l (approved):
lgtm

collivier (approved):

PR #2455, opened 2026-08-24
One name per thing: kebab-case arguments, prefixed environment variables

martin-mat:
The CLI had several names for the same thing and several styles for naming things: kebab-case `cnf-config` next to snake_case `input_config`, an undocumented `cnf-path` alias, `LOGLEVEL`/`LOG_LEVEL` **and** `LOGPATH`/`LOG_PATH` both honored, an unprefixed lowercase `force_install` env var, and `CRYSTAL_ENV=TEST` — a variable named after the compiler — silently switching test-mode behavior.

**One name per thing. Clean breaks, no aliases** — per the release plan: the next release already changes exit-code semantics, and all backward-incompatible changes ship together in it, announced once. The deprecation-alias cycle sketched in #2435 is deliberately dropped.

## Renames

| was | is |
|---|---|
| `cnf-path=` (undocumented alias) | `cnf-config=` |
| `input_config=` / `output_config=` | `input-config=` / `output-config=` |
| `pod_labels=` / `baseline_count=` | `pod-labels=` / `baseline-count=` |
| `force_install` (env) | `CNF_TESTSUITE_FORCE_INSTALL` |
| `CRYSTAL_ENV=TEST` | `CNF_TESTSUITE_ENV=TEST` |
| `LOG_LEVEL` / `LOG_PATH` (and the undocumented `LOGLEVEL`/`LOGPATH` duplicates) | `CNF_TESTSUITE_LOG_LEVEL` / `CNF_TESTSUITE_LOG_PATH` |

## Why the breaks are safe to make

Every retired spelling **fails fast instead of silently doing nothing**: #2423's validation rejects unknown arguments with exit 64 and the list of known ones —

```
$ cnf-testsuite cnf_install cnf-path=x
Unknown argument 'cnf-path='. Known arguments: cnf-config, timeout, input-config, …
$ echo $?
64
```

— and help + completion updated **automatically**, because validation, help, and completion all read the same registry (#2429/#2430's design paying off: this PR touches the allowlist once and never edits the help or completion text for the arg lists).

[...]

PR #2456, opened 2026-08-24
Retry transient network failures on external fetches

### LINKED ISSUE #2445: [MAINTENANCE] Transient network failures fail whole spec shards: no retry around external chart/binary fetches
Every network fetch the suite performs — helm chart pulls, helm repo adds, remote-chart installs, binary downloads — is a **single attempt**. One transient failure from a host we do not control fails the operation, which fails the spec, which reds the whole shard. The failure carries no information about the code under test, so the only response is a manual re-run.

This is currently one of the largest sources of red builds.

### Evidence

Three incidents, three different hosts, within a couple of days:

| host | failure | effect |
|---|---|---|
| `cncf.gitlab.io` | `Get "https://cncf.gitlab.io/stable/coredns-1.13.8.tgz": EOF` | `capabilities` shard red ([run](https://github.com/lfn-cnti/testsuite/actions/runs/32381616476/job/96479951812)) |
| `registry-1.docker.io` | DNS `i/o timeout` during fluentd-bitnami install | collector never installed → `routed_logs` SKIPPED instead of PASSED |
| `dl.k8s.io` | SSL error fetching kubectl | `zombie` shard red |

The first is the clearest demonstration that these are transient rather than real: **the identical command succeeded three times earlier in the same job.** `cnf_install cnf-config=./sample-cnfs/sample-coredns-cnf` completed at 15:26:56, 15:27:40 and 15:28:15, then the 15:28:53 attempt got an EOF. Same chart, same URL, same runner, minutes apart. A single retry would have absorbed it.

### Where

`src/modules/helm/helm.cr` — none of these retry:

- `Helm.pull` (line 368) and `Helm.pull_oci` (387)
- `Helm.install` (338), when the chart is a remote reference
- `Helm.helm_repo_add` (263)

Each ends in `ShellCMD.raise_exc_on_error { ShellCMD.run(cmd, logger) }` — one shot, and any non-zero exit raises.

Also single-attempt: `Halite.get` in `src/modules/tar/tar.cr:109` and `HTTP::Client.get` in `src/tasks/utils/utils.cr:406`.

### Exposure

Hosts reachable at runtime from `src/`, `embedded_files/` and the CNF fixtures, by reference count:

```
 24  raw.githubusercontent.com     6  charts.bitnami.com      2  jaegertracing.github.io
 23  github.com                    4  storage.googleapis.com  2  fluent.github.io
 10  cncf.gitlab.io                3  helm.elastic.co         2  open-policy-agent.github.io
```

Plus `registry-1.docker.io` and `dl.k8s.io` at install time. Nearly every spec that installs a CNF depends on at least one of these answering correctly at that instant.

### Proposal

A small retry wrapper for **network-bound external fetches**, applied to the call sites above.

Important: **retry only what looks transient.** Connection resets/EOF, DNS and connect timeouts, TLS handshake errors, and 5xx should be retried; `chart not found`, `401`, and manifest/validation errors must fail immediately. Blindly retrying everything makes genuine failures take N× longer to report and masks a missing chart as a network problem.

Suggested shape, consistent with what the repo already does:

- Bounded attempts with backoff (3 attempts / 2s, 4s is likely enough — all three incidents were single-attempt blips).
- Tunable by env var following the existing `CNF_TESTSUITE_*` convention in `src/tasks/utils/timeouts.cr`, e.g. `CNF_TESTSUITE_NETWORK_RETRY_ATTEMPTS`, defaulting to something sane for local use.
- Log each retry at WARN with the attempt number and the error, so a flake is visible in the log rather than silently smoothed over. If retries are exhausted, the final error should say how many attempts were made.

The repo already has both halves of the pattern — `repeat_with_timeout` (`src/tasks/utils/timeouts.cr:10`) and `ClusterTools.local_match_by_image_name_with_retries` — they have simply never been applied to network fetches.

### Related

While reviewing the runner state in the run above, `docker ps` showed a kind cluster `9099eea3-…` **"Up 9 days"**, alongside `free5gc-dev` clusters. That is a leaked cluster from a long-finished run holding resources on `unhrunnertwo` — the same host that produced the container-name collisions fixed in #2442. Worth a separate cleanup pass; noting it here so it is not lost.

martin-mat:
Every network fetch the suite performs was a **single attempt** — one blip from `cncf.gitlab.io`, `registry-1.docker.io` or `dl.k8s.io` failed the operation, the spec, and the whole shard, saying nothing about the code under test. #2445 documents three incidents within days; in one, the *identical* chart pull had succeeded three times in the same job minutes earlier.

## Design

**`NetRetry`** ([src/modules/net_retry](src/modules/net_retry/net_retry.cr)) — bounded attempts (3), linear backoff (2s, 4s), both tunable via `CNF_TESTSUITE_NETWORK_RETRY_ATTEMPTS` / `_BACKOFF` per the existing env-var convention. Each retry logs at WARN with attempt number and error, so a flake stays **visible** in the log rather than silently smoothed over; exhaustion logs the count and re-raises the last error **unchanged**, so existing rescue sites keep working.

**Only transient-looking failures retry** — the issue's central caution. Decided two ways:
- by class: socket/TLS-level errors and a `NetRetry::TransientError` (raised by `download_file` for 5xx) — but **not** `File::Error`, which shares `IO::Error` and must not be retried
- by a deliberately narrow message-pattern list: resets, `i/o timeout`, DNS (`no such host`, `name resolution`), `tls handshake`, `EOF`, 5xx status text

`chart not found`, `401`, `404`, values/validation errors → raise on attempt one.

## Applied to (the sites the issue names)

| site | note |
|---|---|
| `download_file` | retry wraps the whole redirect chain; 5xx now classified transient |
| `Helm.helm_repo_add` | `RepoNotFound` stays immediate (permanent) |
| `Helm.install` | remote chart refs are fetched by this command; the filter keeps chart/values errors immediate |
| `Helm.pull` / `Helm.pull_oci` | |
| `TarClient.tar_file_by_url` (Halite) | |

## Verification (Crystal 1.6.2)

- New `net_retry_spec.cr` (tag `points`, 4 examples): classifier both ways (7 transient / 4 permanent messages, class checks incl. the `File::Error` exclusion); succeed-on-attempt-3; exhaust-at-3 and re-raise; permanent = exactly 1 call; **`download_file` end to end** against a local server — recovers after two 503s, fails a 404 on the first attempt
- `crystal spec --tag points` — 53/53
- Live run: `cnf_install` of the coredns sample (real `helm repo add` + remote install through the wrapped paths) — clean

Not addressed: the leaked kind cluster in the issue's "Related" note — that's runner hygiene, not code.

Small doc-merge note: the USAGE.md env-var addition will trivially conflict with #2455's "Other environment variables" section; whichever merges second rebases in seconds.

Closes #2445

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -474,6 +474,17 @@ CNF_TESTSUITE_NODE_DRAIN_TOTAL_CHAOS_DURATION=90
 CNF_TESTSUITE_LABEL_RESOURCE_SLEEP=5
 ```
 
+#### Network retries:
+
+Network-bound fetches (chart pulls, repo adds, binary downloads) are retried on transient
+failures — resets, timeouts, DNS blips, 5xx — while a missing chart or auth error still fails
+immediately:
+
+```
+CNF_TESTSUITE_NETWORK_RETRY_ATTEMPTS=3   # attempts per fetch
+CNF_TESTSUITE_NETWORK_RETRY_BACKOFF=2    # base seconds between attempts (attempt N waits N x backoff)
+```
+
 #### Running The Linter
 
 Ameba (https://github.com/crystal-ameba/ameba) is a static code linter for crystal-lang.
--- a/spec/utils/net_retry_spec.cr
+++ b/spec/utils/net_retry_spec.cr
@@ -1,0 +1,115 @@
+require "../spec_helper"
+require "http/server"
+
+# In-process: exercises the wrapper and download_file directly, with backoff
+# zeroed so exhaustion paths stay fast.
+private def without_backoff(&)
+  ENV[NetRetry::BACKOFF_ENV] = "0"
+  yield
+ensure
+  ENV.delete(NetRetry::BACKOFF_ENV)
+  ENV.delete(NetRetry::ATTEMPTS_ENV)
+end
+
+private def serve(&handler : HTTP::Server::Context, Int32 -> Nil) : {HTTP::Server, String}
+  hits = 0
+  server = HTTP::Server.new do |context|
+    hits += 1
+    handler.call(context, hits)
+  end
+  address = server.bind_tcp("127.0.0.1", 0)
+  spawn { server.listen }
+  Fiber.yield
+  {server, "http://#{address}"}
+end
+
+describe "NetRetry" do
+  it "classifies transient and permanent failures", tags: ["points"] do
+    ["connection reset by peer", "dial tcp: lookup x: i/o timeout", "unexpected EOF",
+     "TLS handshake timeout", "Temporary failure in name resolution",
+     "Unsuccessful request, status code: [503], msg: Service Unavailable",
+     "server responded with 502 Bad Gateway"].each do |message|
+      NetRetry.transient?(Exception.new(message)).should be_true
+    end
+
+    ["chart \"nope\" not found in https://example.com repository",
+     "401 Unauthorized", "Error: INSTALLATION FAILED: values don't meet the specifications",
+     "Unsuccessful request, status code: [404], msg: Not Found"].each do |message|
+      NetRetry.transient?(Exception.new(message)).should be_false
+    end
+
+    NetRetry.transient?(NetRetry::TransientError.new("anything")).should be_true
+    NetRetry.transient?(IO::TimeoutError.new("read timed out")).should be_true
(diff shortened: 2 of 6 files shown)


rich-l (approved):
lgtm

collivier (approved):

PR #2467, opened 2026-08-24
Kubescape 3.0.30 -> 4.0.12 and the NSA framework regolibrary 1.0.316 -> 2.0.33

martin-mat:
## Summary
Phase 4 of the dependency refresh.

- `KUBESCAPE_VERSION` 3.0.30 → 4.0.12, `KUBESCAPE_FRAMEWORK_VERSION` 1.0.316 → 2.0.33.
- v4 release assets are `kubescape_<ver>_<os>_<arch>.tar.gz` (the raw binary is 267 MB vs a 109 MB tarball), so `install_kubescape` downloads and untars instead of download + chmod. `KUBESCAPE_TARGET_BINARY_NAME` is gone; `KUBESCAPE_BINARY` is the single path used by setup and `Kubescape.scan`.
- No parser changes needed: `--format-version=v1` still produces the `controlReports/ruleReports/ruleResponses/alertObject.k8sApiObjects` structure (v4 logs it as deprecated; migrating to v2 is a follow-up). All 16 NSA control names the tests match on are unchanged in regolibrary 2.0.33; C-0048 (HostPath mount) is still scanned by ID.
- **One rule changed what it reports**: C-0034 (`service_account_mapping`) now alerts on the *workloads* that will mount a token (pod-level `automountServiceAccountToken` overrides the SA's) instead of on `ServiceAccount` objects. The test filtered findings to the CNF's ServiceAccounts, so it could never fail under the new rule (caught by the `security` CI shard). It now filters on the CNF's workloads like every other kubescape-backed test.

## Test plan
- [x] `crystal build src/cnf-testsuite.cr`
- [x] kind cluster, kubescape 4.0.12 + regolibrary 2.0.33: `setup:kubescape_scan` and all 14 workload tests (`privilege_escalation` … `hostpath_mounts`) plus `platform:control_plane_hardening`, `platform:cluster_admin`, `platform:verify_secrets_encryption` yield verdicts with no errors
- [x] Spec fixtures behave as specs expect: `sample-nonroot-containers`/`sample-privilege-escalation` (privilege_escalation pass/fail), `sample-nonroot`/`sample-coredns-cnf` (non_root_containers pass/fail), `sample-immutable-fs`/`sample-coredns-cnf` (immutable_file_systems pass/fail), coredns fails `linux_hardening` and `ingress_egress_blocked` as specced
- [x] `service_account_mapping`: `sample-service-accounts` FAILS naming `Pod/nginx` (exit 1); in `sample_multiple_deployments` only the deployment without `automountServiceAccountToken: false` is flagged
- [x] Cross-checked one suspicious result against kubescape 3.0.30 + regolibrary 1.0.316 offline: identical verdict (not a v4 regression)
- [ ] CI shards: `security`, `privileges`, `capabilities`, `platform_security`, and the other kubescape-backed tags

### DIFF
--- a/src/tasks/setup/constants.cr
+++ b/src/tasks/setup/constants.cr
@@ -15,28 +15,15 @@ module Setup
     {% end %}
   end
 
-  KUBESCAPE_TARGET_BINARY_NAME = begin
-    case {TARGET_OS, TARGET_ARCH}
-    when {"darwin", "arm64"}
-      "kubescape-arm64-macos-latest"
-    when {"darwin", "amd64"}
-      "kubescape-macos-latest"
-    when {"linux", "arm64"}
-      "kubescape-arm64-ubuntu-latest"
-    else
-      "kubescape-ubuntu-latest"
-    end
-  end
-
   # Versions of the tools
   # renovate: datasource=github-releases depName=kubernetes-sigs/cluster-api
   CLUSTER_API_VERSION         = "1.9.6"
   # renovate: datasource=github-releases depName=kubernetes-sigs/kind
   KIND_VERSION                = "0.27.0"
   # renovate: datasource=github-releases depName=kubescape/kubescape
-  KUBESCAPE_VERSION           = "3.0.30"
+  KUBESCAPE_VERSION           = "4.0.12"
   # renovate: datasource=github-releases depName=kubescape/regolibrary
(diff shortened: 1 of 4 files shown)


rich-l (approved):
lgtm

collivier (approved):

PR #2468, opened 2026-08-24
Build with Crystal 1.19.1 everywhere, from the official alpine image

martin-mat:
## Summary
Phase 5 of the dependency refresh.

- **One Crystal version, 1.19.1**, in `.crystal-version`, `actions.yml`, `arm64.yml` (was floating), `shard.yml` (`>= 1.19.0`), `SOURCE_INSTALL.md`. Debian/arm64 jobs already built with 1.19.1.
- **Release job uses `crystallang/crystal:1.19.1-alpine-build`** instead of the custom `conformance/crystal:1.6.2-alpine`. The official `-alpine-build` variant already ships every static lib the custom image added, so the build command is unchanged and `tools/crystal-alpine.dockerfile` is deleted.
- **Shards**: `commander` and `icr` removed (unused anywhere; drops `popcorn`/`readline` transitives). `crinja ~> 0.9.0`, `protobuf ~> 2.3.1` pinned (same resolved versions). `sam` stays on the `vulk/sam.cr` fork — the `~task` skip feature only exists there and the fork already contains all upstream commits. `totem`/`halite`/`retriable` are already latest.

## Test plan
- [x] `crystallang/crystal:1.19.1-alpine-build`: `shards install && crystal build --release --static --link-flags '-lxml2 -llzma'` → working 38 MB static binary, no extra packages
- [x] Same image: `crystal spec --tag points` (48 examples) against a kind cluster — all pass; `cli_completion_spec` passes once `bash` is present (alpine has none; CI runners do)
- [x] For reference, 1.21.0-alpine-build also builds the static binary cleanly
- [ ] CI spec matrix on 1.19.1, `release.yml` on the next tag


rich-l (approved):
lgtm

collivier (approved):

PR #2470, opened 2026-08-24
Refresh the fluentd and fluent-bit collectors, drop the Bitnami fluentd flavor

martin-mat:
## Summary
Phase 6.3 of the dependency refresh.

- **fluent-bit** `latest@7d7272…` → `5.1.1@a941bd…`; **fluentd** `v1.19.1-debian-elasticsearch8-1.0` → `v1.19.3-debian-elasticsearch8-1.1@88da01…`. `routed_logs` fingerprints collectors by digest, so these are the values that matter.
- fluent-bit values now pull from Docker Hub (`fluent/fluent-bit`) instead of `cr.fluentbit.io`: the two registries publish **different index digests** for the same tag, so a collector installed from cr.fluentbit.io would not have matched the code's `fluent/fluent-bit@…` fingerprint.
- fluentd values drop the `image.digest` key — the fluentd chart has no such value and silently ignored it; the tag pins the image.
- **Bitnami flavor removed** (`FluentDBitnami`, `install_/uninstall_fluentdbitnami`, its values file, macro, cleanup entry, spec): the chart repo is Broadcom's frozen legacy snapshot, the image is in the discontinued `bitnamilegacy` namespace pinned to an immovable `latest`, and it exercised the same code path as the fluentd flavor.
- The "should fail" spec no longer depends on the Bitnami chart: it installs the official fluentd chart with `spec/fixtures/fluentd-values-bad.yml` (same image as the testsuite flavor, tails `/tmp` instead of container logs, keeps the prometheus source the chart's probes need).

## Test plan
- [x] `crystal build src/cnf-testsuite.cr`
- [x] kind cluster: `setup:install_fluentd` → pod imageID `…@sha256:88da01…`, `routed_logs` PASSED; `setup:install_fluentbit` → `…@sha256:a941bd…`, `routed_logs` PASSED
- [x] bad fluentd from the fixture: daemonset rolls out (probes OK), `routed_logs` FAILED "not being captured" (not SKIPPED — i.e. detected as a collector)
- [ ] CI shard `observability_routed_logs`


rich-l (approved):
lgtm

collivier (approved):

PR #2473, opened 2026-08-26
CI: fail on HTTP errors and retry transient ones in every raw curl

martin-mat:
The `debian`/`arm64` workflows fetch the Crystal tarball, kubectl, and the helm install script with bare `curl -LO` — no `-f`, no `--retry`. Yesterday that cost a red **main**: GitHub's release CDN answered a **54 KB error page** instead of the ~50 MB Crystal tarball, curl saved it happily, `tar` failed on "not in gzip format", and the job died two commands later with the misleading `shards: command not found` ([run 32955880246](https://github.com/lfn-cnti/testsuite/actions/runs/32955880246), `debian (forky)` — the identical leg passed six minutes earlier).

#2456 gave the *suite's* fetches transient-aware retries; these raw workflow curls are the same class one layer up.

## What changes

Every raw curl (9 call sites: `debian.yml` ×4, `arm64.yml` ×5 incl. the nested `stable.txt` lookups) now runs with:

```
curl -fsSL(-O) --retry 3 --retry-delay 2
```

- `-f`: an HTTP error fails the step **at the curl**, with an honest message — never blessing an error page as an artifact
- `--retry`: retries what curl classifies as transient (408/429/5xx, timeouts) — a hard 404 still fails immediately (verified: exit 22 in 0.3s, no retry), so genuine errors aren't slowed down

**`curl | bash` → download-then-execute** for the helm installs: with a pipe, a failed curl feeds bash an empty script, which exits 0 and lets the chain continue as if helm had installed — the same silent-blessing bug in a different costume.

The `install-kind` action already had `-fsSL` and gains the retry for consistency.

## Verification

- YAML parse on all three files
- Flag-combo smoke tests: real fetch works; 404 → exit 22 in 0.3s with no retry; tarball URL answers 200 with the full content-length
- The workflows themselves run on this PR (`debian`/`arm64` trigger on `pull_request`), which is the real test
**Amended after the bullseye failure on [run 32972360467](https://github.com/lfn-cnti/testsuite/actions/runs/32972360467):** a Debian mirror reset mid-`apt install` — a different transient, same misleading `shards: command not found` symptom. Three additions:

- `-o Acquire::Retries=3` on every `apt` call in both workflows
- debian.yml's `&&` chain ended at `git config … safe.directory`, so the final `(cd testsuite && shards install …)` group ran after *any* setup failure and masked the real cause with exit 127; the chain now covers the whole script, as arm64.yml's already did
- `fail-fast: false` on the debian matrix — one release's transient failure no longer cancels the other four

### DIFF
--- a/.github/actions/install-kind/action.yml
+++ b/.github/actions/install-kind/action.yml
@@ -20,7 +20,7 @@ runs:
         esac
         url="https://kind.sigs.k8s.io/dl/${{ inputs.version }}/kind-linux-${arch}"
         if command -v curl >/dev/null 2>&1; then
-          curl -fsSLo ./kind "$url"
+          curl -fsSLo ./kind --retry 3 --retry-delay 2 "$url"
         else
           wget -qO ./kind "$url"
         fi
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -46,9 +46,9 @@ jobs:
         with:
           crystal: 1.19.1
       - run: |
-          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/arm64/kubectl"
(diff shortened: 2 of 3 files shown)


collivier (approved):
seems a bit overkilled but nothing wrong here.

PR #2477, opened 2026-08-26
CI: create kind clusters through one action that retries a missed boot deadline

martin-mat:
kind gives the node containers a fixed 60 s to reach systemd's multi-user target (`could not find a log line that matches "Reached target .*Multi-User System.*"`) — independent of `--wait`, which only governs the later readiness phase. When several jobs of the fan-out boot clusters on the same host at once, that deadline is occasionally missed and the job dies before any test runs. Seen on [run 32971270083](https://github.com/lfn-cnti/testsuite/actions/runs/32971270083): two of four slots on one host timed out at the same second, then booted fine 90 s later.

A new `.github/actions/create-kind-cluster` composite action (sibling of `install-kind`) holds the single create-with-retry sequence: on failure it waits 60 s for the burst to pass, tries once more, and fails with an `::error` if the second attempt also fails. **All five** cluster creations in `actions.yml` now go through it — `spec`, `chaos` and the three `test_binary_*` jobs, the latter of which previously had no retry at all. Each job publishes `CLUSTER` to `$GITHUB_ENV` next to the existing `cluster.env` so the action can be handed the name; `cluster.env` and every later `source cluster.env` are unchanged.

`--retain` is dropped (review feedback): nothing in the workflow inspected retained nodes before the `always()` `Delete Cluster` step removed them, so it only ever served interactive debugging on a runner — and without it kind cleans up its own failed attempt, which removes the need for an explicit delete before the retry. If a boot failure ever needs a post-mortem again, that means re-adding `--retain` *plus* a log-collection step before cleanup.

The action's script was exercised against a fake `kind` for all three paths (fail-then-succeed → 0, fail twice → 1, first-try success → silent). `debian.yml`/`arm64.yml`/`nightly_free5gc.yml` each boot a single cluster on a dedicated runner, so they're deliberately left alone.

### DIFF
--- a/.github/actions/create-kind-cluster/action.yml
+++ b/.github/actions/create-kind-cluster/action.yml
@@ -1,0 +1,38 @@
+name: "Create kind cluster"
+description: "Create a kind cluster from a config file, retrying once when the node containers miss kind's fixed boot deadline. This action is the single place that creates kind clusters in actions.yml."
+inputs:
+  name:
+    description: "Cluster name"
+    required: true
+  config:
+    description: "kind config file"
+    required: false
+    default: "/tmp/cluster.yml"
+  kubeconfig:
+    description: "Where to write the kubeconfig"
+    required: true
+  wait:
+    description: "How long kind waits for the control plane to become ready"
+    required: false
+    default: "5m"
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: |
+        set -euo pipefail
+        cat "${{ inputs.config }}"
+        kind --version
+        # kind gives the node containers a fixed 60 s to boot systemd; when several
+        # jobs boot clusters on one host at once that deadline is occasionally
+        # missed, and a host that busy can also leave the control plane too slow
+        # for kubeadm. kind removes its own failed nodes, so retrying is just
+        # trying again -- after a pause long enough for the burst to pass.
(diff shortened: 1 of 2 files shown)


collivier (changes requested):
I would prefer we remove --retain which is no longer needed.
It was useful when debugging the crashes due to disk speed.

martin-mat:
@collivier agreed — dropped in 32b73bbe. Nothing in the workflow ever looked at retained nodes before the `always()` cleanup removed them, so it only served interactive debugging. Without `--retain` kind removes its own failed attempt, which also let the explicit `kind delete` before the retry go; the action is now just try → warn → wait 15 s → try again → fail loudly. PR description updated with the same reasoning.

martin-mat:
First real exercise of the retry on this PR's own run ([job 98208625911](https://github.com/lfn-cnti/testsuite/actions/runs/32978202713/job/98208625911), `privileges`): attempt 1 missed the 60 s boot deadline, kind cleaned up on its own, attempt 2 booted in 46 s and then lost the control plane to the same host burst (`wait-control-plane: context deadline exceeded` — the API server never answered). The same slot booted its next cluster fine two minutes later, so the pause between attempts is now 60 s instead of 15 s; nothing changes on the happy path.

collivier (approved):

PR #2519, opened 2026-08-28
github-runner image: OCI labels, Renovate annotations, overridable image

### LINKED ISSUE #2511: [MAINTENANCE] Recreate the self-hosted runners from the Crystal 1.19.1 image (they still run 1.6.2)
**What**

The self-hosted runner containers still compile with **Crystal 1.6.2**. Every self-hosted `Crystal Specs (<tag>)` and `Test Binary Without Source(...)` job shows `/crystal/crystal-1.6.2-1/share/crystal/...` in stack traces and `Shard "crinja" may be incompatible with Crystal 1.6.2` in `shards install`, while `.crystal-version` pins 1.19.1 (#2468), the GitHub-hosted chaos job installs 1.19.1, and `tools/github-runner/Dockerfile` has pinned `CRYSTAL_VERSION=1.19.1` since #2471.

The image was rebuilt; the running containers were never recreated from it (`tools/github-runner/create_runners.sh`). Until they are, anything that needs Crystal ≥ 1.19 breaks self-hosted CI silently — e.g. `Time.instant` (not in 1.19.1 either, but the point stands for every stdlib change since 1.6), and the 1.6.2 toolchain is what has actually been validating every PR on the self-hosted shards.

**Action (host operators)**

- Recreate every runner container (`unhrunner*`, `unhrunnertwo*`, `unhrunnerthree*`, `unhrunnerfour*`, `unhrunnerfive*`) from the rebuilt image.
- Verify with any self-hosted job: `crystal --version` in the build step, or the absence of the 1.6.2 path in a stack trace.

**Also observed on the pool (2026-08-26/27), for the same hands**

- Host **five** (`unhrunnerfive0–3`) was the slow one: three of four `Create Kind Cluster` failures that day (boot deadline / `wait-control-plane`) were on its slots, one within seconds of a sibling succeeding; #2477's retry absorbs this class but does not fix the host.
- A job's `Delete Cluster` step on `unhrunnerfive1` logged `could not kill container … did not receive an exit event` for a kind node (12:24 that day); the container was gone by the next job, so no leak was found — worth keeping an eye on.
- The pool saturates when several PRs push at once: runs sat `queued` with 40–60 shards pending for up to ~90 min and three shards were cancelled before ever getting a runner.

Closing #2158 pointed here.

martin-mat:
Phase 1, step 4 of the container-image plan. The runner image stays (installing Crystal/kind/kustomize per shard would cost ~30 s × ~50 shards per run); this only prepares it for the images workflow.

- OCI labels (`source` links the future GHCR package to this repo).
- `# renovate:` annotations for the `myoung34/github-runner` base image, kind and kustomize (`kustomize/vX.Y.Z` tags need a regex versioning); Crystal already had one.
- `create_runners.sh` honours `RUNNER_IMAGE` from the environment, default unchanged.

### Verification

Built locally and smoke-tested: Crystal 1.19.1, Shards 0.20.0, kind 0.32.0, kustomize v5.8.1, kubectl v1.36.4, helm v4.2.4, gh 2.98.0. Not deployed — the redeploy from GHCR is phase 3 and closes #2511.

### DIFF
--- a/tools/github-runner/Dockerfile
+++ b/tools/github-runner/Dockerfile
@@ -1,10 +1,21 @@
+# Self-hosted CI runner with the toolchain baked in, so that no shard pays
+# for installing Crystal, kind or kustomize. Published by the images workflow
+# as ghcr.io/lfn-cnti/github-runner:<runner version>-crystal<Crystal version>.
+
+# renovate: datasource=docker depName=myoung34/github-runner
 FROM myoung34/github-runner:2.336.0-ubuntu-noble
 
 # renovate: datasource=github-releases depName=crystal-lang/crystal
 ARG CRYSTAL_VERSION=1.19.1
 ARG CRYSTAL_URL=https://github.com/crystal-lang/crystal/releases/download
+# renovate: datasource=github-releases depName=kubernetes-sigs/kind
 ARG KIND_VERSION=v0.32.0
+# renovate: datasource=github-releases depName=kubernetes-sigs/kustomize versioning=regex:^kustomize/v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)$
 ARG KUSTOMIZE_VERSION=v5.8.1
+
+LABEL org.opencontainers.image.source="https://github.com/lfn-cnti/testsuite" \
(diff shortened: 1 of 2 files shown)


collivier (approved):

collivier (changes requested):
I would rather remove this runner. we should use an upstream one and install crystal as first workflow step. That would ease switching to public one if needed.

Issue #2153, opened 2024-09-19
[Feature] CNF Installation (5), Prepare tests for new installation method


kosstennbl:
Some of the tests in testsuite are not compatible with planned new installation method. They should be redesigned or given skip conditions if not suitable for some situations.

Potential issues:
- Tests that are specific to install method
- Tests that don't support multiple namespaces
- Tests that use/replicate current install methods
- Tests that use configmaps, created outside of the test

List of problematic tests (possibly incomplete) and ideas for problem solutions:
- 5g_validator.cr: Not found
-  compatibility.cr: 
    - "helm_chart_valid", "helm_chart_published", "helm_deploy": Check only through helm_charts and helm_dirs deployment types (where applicable).
    - "cni_compatible": Check if cnf_to_new_cluster process is suitable for new installation method. Ensure correct test logic.
-  configuration.cr:
    - Kyverno and its policies should work fine, they are applied cluster-wide
    - "ip_addresses" should be removed, "hardcoded_ip_addresses_in_k8s_runtime_configuration" should look through common manifest
    - "alpha_k8s_apis" cnf_to_new_cluster usage, needs checking. 
- microservice.cr:
    - "reasonable_startup_time": uses configmap that is created on installation.
    - "service_discovery": dependent on using single namespace for resource kind name discovery. 
- observability.cr:
    - "prometheus_traffic", "open_metrics": uses release_name to create configmap during test. Configmap should be renamed.
    - "tracing": uses configmap that is created on installation.
- ran.cr:
    - "oran_e2_connection": uses configmap that is created on installation.
- reliability.cr:
    - Litmus tests don't seem to create problems in the scope of this epic. Paths for writing experiment results should be changed.
- security.cr:
    - Kubescape scans are done cluster-wide, should be ok.

kosstennbl:
Additional issue was discovered: we need to somehow ensure correct namespaces for each resource for cnf_workload_resource method. If deployments in CNF would be on different namespaces, get_deployment_namespace method wouldn't work.

kosstennbl:
Done in #2163 #2164 #2166 #2167

PR #2531, opened 2026-09-01
Remove the RAN and 5G test suites and everything only they used

martin-mat:
Removes the RAN and 5G test suites and everything that existed only for them (draft; CI is the arbiter).

**Tests removed**: `oran_e2_connection` (disabled since the v2 install rework, #2153/#2107, and dependent on the unrebuildable `conformance/oran`/`conformance/flexric` images), `smf_upf_core_validator`, `smf_upf_heartbeat`, `suci_enabled`, with the `ran` and `5g` categories/suites and their entries in `cnf-testsuite help`.

**Removed with them**:
- `ORANMonitor`, the never-called `SRSRAN` helper, UERANSIM tooling (`ueransim.cr`, `setup:install_ueransim`/`uninstall_ueransim`, the `ues-values` template and embedded UE config, `FIVE_G_TOOLS_DIR`), the `K8sTshark` capture helper (its last consumers), and the `--baseline-count` option only the heartbeat test read
- the `five_g_parameters` config section including `ric_label` — v2 (`FiveGParameters`), v1 getters, and the v1→v2 migration — with its documentation in `CNF_TESTSUITE_YML_USAGE.md`
- fixtures `sample-oran-ric`, `sample-oran-noric`, `sample_srsran_ueauth_open5gs`, `sample_open5gs`, `sample_open5gs_no_auth` and `spec/5g/`
- the `oran` tag in the CI chaos job (renamed *Chaos Tests*) and the `5g`/`core` tags in the exclusion list; the *RAN Tests* and *5G Tests* doc sections and TOC entries

**Kept on purpose**: free5GC as the reference CNF (`example-cnfs/free5gc`, the nightly workflow, `spec/free5gc_cert_spec.cr`) — its nightly runs the certification set, which uses none of the above; its config never set `five_g_parameters`.

Note for users: a v2 config that still carries `five_g_parameters:` (or `ric_label`) will be rejected by the strict parser — it was only ever read by the removed tests.

### Verification

Built with the CI compiler and the full spec graph compiles (no examples executed); a straggler grep over `src/`, `spec/` and the docs finds no remaining reference; `mdl` passes. CI is the arbiter for the rest — the `oran` chaos shard simply disappears.


collivier (approved):
Same for Open5gs and UERANSIM, right ? or do we keep there ?

PR #2532, opened 2026-09-01
Remove the 5G test suite and everything only it used

martin-mat:
Stacked on #2531 (merge that first). Removes the 5G suite and everything that existed only for it:

- tests `smf_upf_core_validator`, `smf_upf_heartbeat`, `suci_enabled` and the `5g` category/suite; the `--baseline-count` option only the heartbeat test read
- UERANSIM tooling (`ueransim.cr`, `setup:install_ueransim`/`uninstall_ueransim`, the `ues-values` template and embedded UE config, `FIVE_G_TOOLS_DIR`) and the `K8sTshark` capture helper (its last consumer)
- the `five_g_parameters` config section — v2 (`FiveGParameters`), v1 getters, and the v1→v2 migration — with its documentation in `CNF_TESTSUITE_YML_USAGE.md`
- fixtures `sample_open5gs`, `sample_open5gs_no_auth` and `spec/5g/`; the `5g`/`core` tags in the CI exclusion list; the *Category: 5G Tests* doc section and TOC entry; the 5G entry in `cnf-testsuite help`

**Kept on purpose**: free5GC as the reference CNF (`example-cnfs/free5gc`, the nightly workflow, `spec/free5gc_cert_spec.cr`) — its nightly runs the certification set, which does not use any of the above; its config never set `five_g_parameters`.

Note for users: a v2 config that still carries a `five_g_parameters:` block will be rejected by the strict parser.

### Verification

Built with the CI compiler and the full spec graph compiles (no examples executed); a straggler grep over `src/`, `spec/` and the docs finds no remaining reference; `mdl` passes. CI is the arbiter for the rest.


collivier (approved):
ok it anwsers my previous question in first PR

martin-mat:
Folded into #2531 as one PR.

PR #2541, opened 2026-09-15
Improve elactic_volumes accuracy and increase its pass/skip/fail test coverage

collivier:
It fixes how elastic_volumes reports results:
- skip the test when the CNF uses no PVC-backed volumes instead of crashing or misreporting.
- treat a PVC that references a nonexistent StorageClass as non-elastic (FAIL) rather than raising a NotFoundError.

It also adds sample-cnfs/sample-elastic-pvc, a minimal CNF whose Deployment mounts a PVC on the elastic default storage class.


martin-mat (commented):
Thanks — evaluating only PVC-backed volumes and skipping volume-less CNFs fixes real false failures, and the pass/skip/fail coverage plus the `sample-elastic-pvc` fixture are a clear improvement. Inline comments below; three points that sit outside the diff:

**Missing PVC still crashes the test (only the StorageClass lookup is guarded).** The `NotFoundError` rescue covers `Get.resource("storageclasses", …)`, but the sibling PVC lookups still raise: `Volume.storage_class_by_volumes` (`state.cr:94`) and the guessed `"<vct>-<name>-0"` fetch in `VolumeClaimTemplate.storage_class_by_vct_resource` (`state.cr:182`). A typo'd `claimName`, a deleted PVC, or a VCT StatefulSet scaled to 0 → NotFoundError → test recorded as `error`, exit 2 — the same crash class this PR targets, one call site over. A shared nil-returning lookup (precedent: `KubectlClient::Get.resource_uid` rescues NotFoundError internally) would fix all three sites uniformly.

**`database_persistence` now diverges from the block fixed here** (`state.cr:442`). It has the same evaluate-elasticity code but didn't get the PVC filter, so its failure reason still lists configMap/secret/emptyDir volume names — the same mysql StatefulSet fails the two tasks with different volume lists. Caveat when unifying: passing raw volumes is also what keeps `database_persistence` correct for VCT StatefulSets (see the inline comment on the gate). One shared helper for evaluation + reason string fixes both.

**Docs** (`docs/TEST_DOCUMENTATION.md` › Elastic volumes): worth a line that CNFs without persistent volumes are now SKIPPED — the neighbouring Database persistence entry documents its skip condition, and a user whose CNF previously FAILED here will otherwise see SKIPPED after upgrade with no explanation.

src/tasks/workload/state.cr
```diff
@@ -367,25 +372,25 @@ scored_task "elastic_volumes",
       Log.for("elastic_volumes:test_resource").debug { resource.inspect }
       Log.for("elastic_volumes:volumes").debug { volumes.inspect }
 
-      next true if volumes.size == 0
+      # Only persistent (PVC-backed) volumes are evaluated for elasticity. ConfigMap,
+      # Secret and emptyDir volumes are not persistent storage and have nothing to check.
+      pvc_volumes = volumes.as_a.select { |volume| volume.dig?("persistentVolumeClaim", "claimName") }
+      next true if pvc_volumes.empty?
```
martin-mat, line 378: **StatefulSets using `spec.volumeClaimTemplates` are now skipped instead of evaluated.** VCT-backed storage never appears as a `persistentVolumeClaim` entry in `spec.template.spec.volumes` (that's why `VolumeClaimTemplate.pvc_name_by_vct_resource` reconstructs the PVC name), so for e.g. the in-repo Bitnami mariadb/postgresql samples — a configMap volume plus volumeClaimTemplates — `pvc_volumes` is empty, this `next true` fires, and the result is SKIPPED "No persistent volumes are used". Pre-PR the resource went through the VCT branch of `WorkloadResource.elastic?`. That's a silent loss of coverage for the most common stateful shape.

Suggested gate: `next true if pvc_volumes.empty? && !VolumeClaimTemplate.vct_resource?(full_resource)` (moving the `full_resource` fetch above it), and a spec on a VCT StatefulSet fixture to lock it in.

src/tasks/workload/state.cr
```diff
@@ -114,8 +114,13 @@ module StorageClass
     #todo elastic_by_storage_class?
     elastic = false
     provisioners = storage_class_names.reduce( [] of String) do |acc, storage_class|
-      resource = KubectlClient::Get.resource("storageclasses", storage_class.dig?("class_name").to_s, namespace)
-      if resource.dig?("provisioner")
+      resource = begin
+        KubectlClient::Get.resource("storageclasses", storage_class.dig?("class_name").to_s, namespace)
+      rescue ex : KubectlClient::ShellCMD::NotFoundError
```
martin-mat, line 119: Two things about resolving a missing StorageClass this way:

1. `elastic_by_storage_class?` is any-provisioner-wins (`elastic = true` on the first match), so dropping the missing class from `provisioners` means a workload with two PVCs — one on an elastic class, one whose `storageClassName` doesn't exist and can never bind — passes "All used volumes are elastic", while this log line says the volume "is not elastic". The single-PVC case works as intended (empty list → fail); only the mixed case slips through.
2. The diagnosis stays at info level: the user sees the generic "uses non-elastic volumes: <name>" reason and goes looking at provisioner drivers, when the real problem is a StorageClass that doesn't exist. Worth surfacing via `result.append_description`/the impacted-resource reason.

src/tasks/workload/state.cr
```diff
@@ -367,25 +372,25 @@ scored_task "elastic_volumes",
       Log.for("elastic_volumes:test_resource").debug { resource.inspect }
       Log.for("elastic_volumes:volumes").debug { volumes.inspect }
 
-      next true if volumes.size == 0
+      # Only persistent (PVC-backed) volumes are evaluated for elasticity. ConfigMap,
+      # Secret and emptyDir volumes are not persistent storage and have nothing to check.
+      pvc_volumes = volumes.as_a.select { |volume| volume.dig?("persistentVolumeClaim", "claimName") }
```
martin-mat, line 377: Nit: this is the same predicate `Volume.storage_class_by_volumes` already applies (`volumes.select { |x| x.dig?("persistentVolumeClaim", "claimName") }`, line 91), so "what counts as persistent" now lives in two places. When the definition widens (inline `ephemeral:` volumes, or the VCT case above) one site will drift. A small shared `Volume.pvc_volumes(volumes)` used by both keeps it in one place.

src/tasks/workload/state.cr
```diff
@@ -367,25 +372,25 @@ scored_task "elastic_volumes",
...
       Log.for("#{t.name}:elastic_result").info {elastic_result}
       unless elastic_result
-        result.add_impacted_resource(resource["kind"], resource["name"], resource["namespace"], reason: "uses non-elastic volumes: #{volumes.as_a.map(&.dig("name")).join(", ")}")
+        result.add_impacted_resource(resource["kind"], resource["name"], resource["namespace"], reason: "uses non-elastic volumes: #{pvc_volumes.map(&.dig("name")).join(", ")}")
       end
     
       elastic_result
     end
 
     Log.for("elastic_volumes:result").info { "Volumes used: #{volumes_used}; Elastic?: #{all_volumes_elastic}" }
```
martin-mat, line 391: Nit: `all_volumes_elastic` is assigned `true` once (line 368) and never reassigned — the verdict comes from `task_response` — so this summary always prints `Elastic?: true`, including right before `result.failed(...)` in the new sample-local-storage spec. Suggest dropping the variable and logging `task_response` here.

spec/workload/state_spec.cr
```diff
@@ -45,13 +43,41 @@ describe "State" do
...
+      ShellCmd.cnf_install("--cnf-config ./sample-cnfs/sample-elastic-pvc/cnti-testsuite.yaml")
+      result = ShellCmd.run_testsuite("elastic_volumes", cmd_prefix: "CNTI_TESTSUITE_LOG_LEVEL=debug")
+      (/(PASSED).*(All used volumes are elastic)/ =~ result[:output]).should_not be_nil
+    ensure
+      result = ShellCmd.cnf_uninstall()
+      result[:status].success?.should be_true
+    end
+  end
+
+  it "'elastic_volumes' should fail if the cnf uses non-elastic persistent volumes", tags: ["elastic_volume"]  do
```
martin-mat, line 68: Nit: this setup/teardown (schedulable-node lookup + the two `update_yml` calls on `worker-node-value.yml`) is a verbatim copy of the `no_local_volume_configuration` spec below. Forgetting the resetting `update_yml` in one `ensure` leaves the fixture dirty and breaks the other example — a small helper wrapping install/teardown of sample-local-storage would hold it once.

Also, none of the new examples cover a StatefulSet with volumeClaimTemplates — one would have caught the skip regression flagged on `state.cr:378`.

src/tasks/workload/state.cr
martin-mat: Thanks — `1e0e4e9` surfaces the missing class in the reason, which covers point 2. Point 1 is still open: `elastic_by_storage_class?` still sets `elastic = true` on the first matching provisioner, so a workload with one elastic PVC and one whose StorageClass doesn't exist passes, and the new "missing storage class(es)" text is only reachable when the verdict is already a fail. A PVC whose class doesn't exist can never bind, so I'd let `missing_classes` decide: `elastic = false` whenever it is non-empty (the return shape already carries it). Then the reason text fires exactly when it should.

martin-mat:
Reviewed `1e0e4e9` against the earlier review — 7 of 8 points are addressed, so I've resolved those threads:

- VCT StatefulSets are evaluated again (gate includes `vct_resource?`), with the new `sample-elastic-vct` fixture and example covering the shape that regressed — and the `elastic_volume` shard is green with it.
- Missing PVCs no longer crash the test (both lookups rescue `NotFoundError`).
- `database_persistence` uses the same PVC filter and reason; the dead `all_volumes_elastic` is gone; the PVC predicate lives once in `Volume.pvc_volumes`; the sample-local-storage ritual is a helper; docs mention the skip.

One thread stays open (`state.cr` — `elastic_by_storage_class?`): the missing StorageClass is now surfaced in the reason but doesn't affect the verdict, which is still any-provisioner-wins. Details in the thread; a one-line change (`elastic = false` when `missing_classes` is non-empty) closes it. Otherwise this looks ready.

martin-mat:
lgtm after addressing the last piece.

PR #2551, opened 2026-09-19
Add cert and workload workflow for example CNFs

collivier:
it runs the CNTi cert and workload suites against each example CNF on its own kind cluster.
The run must exit below 2, cert must print a PASSED verdict, and workload must report a final score.


rich-l (approved):
lgtm

PR #2552, opened 2026-09-19
reasonable_image_size: make limit configurable, skip instead of fail on env errors, demote to bonus

collivier:
The test was weak because it conflated environment problems with CNF defects and leaned on a one-size-fits-all threshold:

- The 5 GB limit was hard-coded for every CNF; the only override was a 16 MB CI hack gated on CNTI_TESTSUITE_ENV=TEST.
- Any docker pull/save/gzip failure or unparseable size hard-failed the test even when the image was perfectly fine. Registry auth, air-gap or mirror registries, transient network errors and multi-arch images with no manifest for the node platform all turned into CNF failures.
- Only Deployment/StatefulSet/Pod/ReplicaSet were inspected; DaemonSets and other workload kinds were skipped silently, and the same image was pulled and measured once per container reference.
- The output was opaque: no measured size was ever reported, so there was nothing to tune.

What changes:

- The limit is now configurable per CNF via image_size_max_mb in the <common> section of cnti-testsuite.yaml and via the CNTI_TESTSUITE_MAX_IMAGE_SIZE_MB environment variable. The default is unchanged at 5000 MB (16 MB kept for CNTI_TESTSUITE_ENV=TEST).
- Images that cannot be measured are warned about and no longer fail the test; the test is skipped when nothing could be measured.
- Every container image is checked (all workload kinds), and duplicates are measured only once.
- The measured size is reported per image, with a warning band at 80% of the limit.
- Demoted from Normal (5 pts) to Bonus (1 pt): with a tunable threshold it is hygiene, not certification material. The points spec is updated accordingly.


rich-l (approved):
lgtm

PR #2553, opened 2026-09-20
cpu_limits: require CPU requests or limits instead of certifying on limits

collivier:
CPU limits are disputed: they throttle CPU usage and add tail latency, and the Pod Security Standards only require requests. Certifying CNFs on a practice many advisories (Borg, GKE Autopilot, CPU-Manager) recommend against overweights a contested rule.

The test no longer delegates to Kubescape's "Ensure CPU limits are set" control. It inspects the workload directly and:

- passes when a container has CPU limits;
- passes with an INFO note when it has CPU requests only (requests-only posture);
- fails only when a container has neither CPU requests nor CPU limits.
- reports each affected container as an impacted resource and keeps a remediation message.

The test is now Bonus (1 pt) instead of Essential (100 pts), matching memory_limits staying Essential for node protection while CPU limits remain hygiene. Dropping the kubescape_scan dependency also makes the test usable on air-gapped clusters.


rich-l (approved):
lgtm

PR #2554, opened 2026-09-20
operator_installed: verify operator health and report N/A when none

collivier:
The test was weak: it returned N/A ("No Operators Found") whenever OLM Subscriptions were absent or a CSV never resolved, so a stuck or broken operator looked the same as a CNF that simply has no operator. When Subscriptions existed it only checked that the CSV reached phase Succeeded and never verified the operator was actually running.

What changes:

- With no Subscription, return N/A with an explicit message ("the CNF is not operator-managed"): the test is informational and non-OLM operators are out of scope.
- A Subscription that never gets an installedCSV, or a CSV that does not reach InstallSucceeded/Succeeded, is now reported as a finding (failed) instead of silently degrading to N/A. Wait timeouts return false and are handled, so a stuck operator no longer crashes the category.
- After a CSV succeeds, the operator Deployments named by the CSV install strategy are waited on and must be ready; any that are not are reported.
- Each issue is appended to the result description.

Kept Bonus: it is an install-pattern check, not a security property. The existing specs still match ("Operator is installed" / "No Operators Found").

References: https://github.com/lfn-cnti/testsuite/issues/2548


rich-l (approved):
lgtm

